PR #68, opened 2019-11-22
provide_a_default_base-voltages.yml :

FranckLecuyer:
Adding base-voltages.yml in resources
Using this resource if no config file is found in the platform config directory (.itools directory)

Signed-off-by: Franck LECUYER <franck.lecuyer@rte-france.com>

**Other information**:


single-line-diagram-color/src/main/java/com/powsybl/sld/color/BaseVoltageColor.java
```diff
@@ -27,13 +27,22 @@
...
 
-    private final BaseVoltagesConfig config;
+    private BaseVoltagesConfig config;
 
     public BaseVoltageColor(Path configFile) throws IOException {
         Objects.requireNonNull(configFile);
         Yaml yaml = new Yaml(new Constructor(BaseVoltagesConfig.class));
-        InputStream configInputStream = Files.newInputStream(configFile);
-        config = yaml.load(configInputStream);
+        try {
```
geofjamg: It would be better to use Files.exist() instead of relying of the exception?

single-line-diagram-color/src/main/resources/base-voltages.yml
```diff
@@ -0,0 +1,18 @@
...
+      maxValue: 500
+      color: "#FF3333"
+      profile: "RTE"
+    - name: "225"
+      minValue: 180
+      maxValue: 300
+      color: "#33FF33"
+      profile: "RTE"
+
+defaultProfile: "RTE"
```
geofjamg: Could you rename it Default?

FranckLecuyer: Is it "RTE" that should be renamed to "Default", or the property name "defaultProfile" ?

geofjamg: RTE

Issue #50, opened 2019-11-06
Compute and add viewbox attribute


jonenst:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
The svgs don't have a viewbox

* **What is the expected behavior?**
The svgs should have a viewbox.

* **What is the motivation / use case for changing the behavior?**
This allows to give them an aspect ratio and preferred size that so that the client doesn't have to render them to know it. This also allows to have elements at negative coordinates that are displayed in viewers (currently only the voltage level name is like this)

jonenst:
depends on #49

PR #87, opened 2019-12-13
[wip] viewbox

jonenst:

#50


jonenst:
Not ready at all.
- the computation is not done for cgmes vl, substation and zone
- the computation for positionVoltagelevelLayout crops the feeder names at the bottom
-  the computation for positionVoltagelevelLayout does not have symetrical padding on left and right (20 and 80)
- Need to write a code that will compute a default size by just making the bounding box of the nodes and call it if the size can not be computed by a more specific algorithm (in case of missing data for cgmes right now)

PR #91, opened 2020-01-10
Smart substation layout algorithm

FranckLecuyer:

**Other information**:


jonenst:
I looked into this. 
- the org.openide.* artifacts from the netbeans repo might become accessible on maven central some day: https://issues.apache.org/jira/browse/INFRA-19735 
- they are not needed in our case so just exclude them
- the fastutil jar is 18Mb and can be reduced to 2Mb using the find-deps.sh but we have to find a way to host it if we do so

jonenst:
Looked into it a bit more. The org.openide artifacts were available on central with version RELEASE90 instead of 82. I used them. Stax-utils was also available on central (different version than gephi's dependency but the content of the jar is them same). stax-utils from central depends on a missing artifact (com.bea.xml jsr173-ri ) due to licensing problems, but we already have the same problem in powsybl-core and simply excluded it.

All in all, all those dependencies are barely used, but having them gives more confidence that it won't throw a LinkageError...

The build now depends on maven central only.

jonenst:
Maybe exclusions are not even needed for the openide artifacts and merely specifying the version that is available on central is enough.

Issue #84, opened 2019-12-09
Exception when generating substation diagram from CGMES-DL


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
Probably because of #10 , CGMES substation layout is broken. 
```java
2019-12-09_17:04:16.527 [Thread-15] INFO  com.powsybl.sld.SubstationDiagram - Writing SVG and JSON metadata files...
2019-12-09_17:04:16.551 [JavaFX Application Thread] ERROR c.p.s.v.a.AbstractSingleLineDiagramViewer - java.lang.AssertionError: One node must be outside any graph
java.lang.AssertionError: One node must be outside any graph
	at com.powsybl.sld.svg.DefaultSVGWriter.drawSnakeLines(DefaultSVGWriter.java:978)
	at com.powsybl.sld.svg.DefaultSVGWriter.drawSubstation(DefaultSVGWriter.java:398)
	at com.powsybl.sld.svg.DefaultSVGWriter.writegraph(DefaultSVGWriter.java:366)
	at com.powsybl.sld.svg.DefaultSVGWriter.write(DefaultSVGWriter.java:327)
	at com.powsybl.sld.SubstationDiagram.writeSvg(SubstationDiagram.java:122)
```

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Using the sample of #5 

```java

Network network = Importers.loadNetwork("sample.zip");

new LayoutToCgmesExtensionsConverter(new VerticalSubstationLayoutFactory(),
                                                                    new PositionVoltageLevelLayoutFactory(new PositionByClustering()), new LayoutParameters(), true)
                .convertLayout(network);

new CgmesDLExporter(network)
                .exportDLData(new FileDataSource(Paths.get("/tmp"), "sample"));
```

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**
  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

PR #92, opened 2020-01-16
Fix CGMES substation layout, aligning it to the latest changes of the…

massimo-ferraro:
… SVG writer

Signed-off-by: massimo.ferraro <massimo.ferraro@techrain.eu>

#84

**Other information**:

### DIFF
--- a/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/main/java/com/powsybl/sld/cgmes/layout/CgmesSubstationLayout.java
+++ b/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/main/java/com/powsybl/sld/cgmes/layout/CgmesSubstationLayout.java
@@ -6,6 +6,8 @@
  */
 package com.powsybl.sld.cgmes.layout;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


massimo-ferraro:
Actually, with this changes, all the substation layouts (horizontal, vertical and CGMES) split the edges in 2 parts. So the assertion error should not be there.
If you want to be sure that also new future substation layout do not raise the error, the splitting could be moved to the graph creation (leaving to the layout the setting of the coordinates of the added fictitious node). Is this what you had in mind?
This would assume that you want to always have this splitting/fictitious node added. 
The other option would be to let the layouts do the splitting, if they need it; and to change the SVG writer for not assuming that the splitting/addition of the node has been done.

FranckLecuyer:
> Actually, with this changes, all the substation layouts (horizontal, vertical and CGMES) split the edges in 2 parts. So the assertion error should not be there.
> If you want to be sure that also new future substation layout do not raise the error, the splitting could be moved to the graph creation (leaving to the layout the setting of the coordinates of the added fictitious node). Is this what you had in mind?
> This would assume that you want to always have this splitting/fictitious node added.
> The other option would be to let the layouts do the splitting, if they need it; and to change the SVG writer for not assuming that the splitting/addition of the node has been done.

FranckLecuyer:
Maybe the best solution is the last one : changing the SVG writer by eliminating the assertion error, and adapt the code to avoid any further exceptions ...

PR #96, opened 2020-02-04
removing_iidm_network_references_in_default_diagram_style_provider

FranckLecuyer:
Removing all iidm class references in DefaultDiagramStyleProvider class (network, three windings transformer, voltage level), by adding the necessary informations in the fictitious 3WT node class, during the graph building phase.

Modification of some of the test cases subsequently.

Signed-off-by: Franck LECUYER <franck.lecuyer@rte-france.com>

**Other information**:

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Fictitious3WTNode.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Fictitious3WTNode.java
@@ -9,6 +9,7 @@ package com.powsybl.sld.model;
 import com.fasterxml.jackson.core.JsonGenerator;
 
 import java.io.IOException;
+import java.util.Map;
 
 import static com.powsybl.sld.library.ComponentTypeName.THREE_WINDINGS_TRANSFORMER;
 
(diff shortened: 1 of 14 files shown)


single-line-diagram-core/src/main/java/com/powsybl/sld/model/Fictitious3WTNode.java
```diff
@@ -17,20 +18,43 @@
  */
 public class Fictitious3WTNode extends FictitiousNode {
 
-    private final String transformerId;
+    private final Map<Feeder3WTNode.Side, String> idsLegs;
```
geofjamg: Isn't redundant with information from VoltageLevelGraph?
A node knows its (node, edge) neighbors, and should be possible to get it by just looking at neighbors.
Also maybe Fictitious3WTNode should not extend FictitiousNode and should have a more specific design
   - only 3 edges possible (so edge1, edge2, edge3 attributes and not a list)
   - these 3 edges are kind of ordered (leg 1, 2, 3)
   - it would allow also to refactor TwtEdge to be more "type safe": one side with a  Fictitious3WTNode and the other with a FeederNode

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramStyleProvider.java
```diff
@@ -221,49 +213,39 @@ public String getIdWireStyle(Edge edge) {
...
 
         if (nameSubComponent.equals(WINDING1)) {
-            if (leg1Vl.getId().equals(vlId)) {
-                otherSide = !rotateSVG ? side3 : side2;
-            } else if (leg2Vl.getId().equals(vlId)) {
-                otherSide = !rotateSVG ? side3 : side1;
-            } else if (leg3Vl.getId().equals(vlId)) {
-                otherSide = !rotateSVG ? side2 : side1;
+            if (idsLegs.get(Feeder3WTNode.Side.ONE).equals(vlId)) {
+                otherSide = !rotateSVG ? Feeder3WTNode.Side.THREE : Feeder3WTNode.Side.TWO;
```
geofjamg: Is it the right place to deal with rotation? This class should only be responsible for style but not for layout.

single-line-diagram-core/src/main/java/com/powsybl/sld/model/Fictitious3WTNode.java
FranckLecuyer: Fictitious3WTNode is created only when we display a voltage level diagram (and not in the case of a substation or zone diagram), so we need to store in that kind of node the ids and nominal voltage of the other voltage levels ...

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramStyleProvider.java
FranckLecuyer: In another issue, I'm going to try to move the svg component rotation outside the default diagram style provider methods (in fact, in the SVG writer before calling the style provider getNodeSVGStyle method), and just use this information for generating the right style ...

PR #126, opened 2020-04-29
add unescapedId attribute in metadata

AbdelHedhili:
Signed-off-by: AbdelHedhili <abdelsalem.hedhili@rte-france.com>

add the attribute "unescapedId" for the switches in metadata

no attribute unescapedId

attribute unescapedId

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
@@ -47,6 +47,8 @@ public class GraphMetadata implements AnchorPointProvider {
 
         private final boolean vLabel;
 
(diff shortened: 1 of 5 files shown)


single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
```diff
@@ -47,6 +47,8 @@
 
         private final boolean vLabel;
 
+        private final String unescapedId;
+
```
FranckLecuyer, line 51: this field correspond to an equipmentId : so maybe we can rename it ?

geofjamg: You are right, equipmentId would be more meaningful

PR #133, opened 2020-06-08
indicate open lines

FranckLecuyer:

Feature

Open lines (on one or both ends) are not graphically distinguished from closed lines.

Open lines (on one or both ends) are graphically distinguished from closed lines, with colored/black plain/dashed line.

**Other information**:

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LayoutParameters.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LayoutParameters.java
@@ -68,6 +68,8 @@ public class LayoutParameters {
     private boolean labelCentered = false;
     private boolean labelDiagonal = false;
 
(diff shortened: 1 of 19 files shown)


single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LayoutParameters.java
```diff
@@ -68,6 +68,8 @@
     private boolean labelCentered = false;
     private boolean labelDiagonal = false;
 
+    private boolean indicateOpenLines = true;
```
geofjamg, line 71: After discussion with @jonenst it seems that highlightLineState would be more correct

FranckLecuyer: Parameter renamed

Issue #15, opened 2019-09-17
Node decorators


geofjamg:
* **Do you want to request a *feature* or report a *bug*?**
Feature

It would be nice to be able to add custom SVG around node like labels (for instance to add a lock picture in case of line planned outage)

cedrtheb:
Hello,
Maybe some pictures to illustrate ?
Regards,
Cédric.

cedrtheb:
@geofjamg  Any news on this ?
What about the possibility to generalize it to any component?
Regards,
Cédric.

PR #142, opened 2020-06-24
Node decorators

floriand-e2r:

Issue #15 

Feature

Issue #15 

Nodes can be decorated with an image

**Does this PR introduce a breaking change or deprecate an API?** yes 
*If yes, check the following:*

**Other information**:


geofjamg:
@floriand-e2r when it is ready to final review, you have to remove the draft status and the WIP

floriand-e2r:
Currently the node decorator position expected is relative to the upper left position of the node image. This was done to have the same behaviour as the label position. But I'm wondering if it really makes sense: one might expect a position relative to the center of the node image. This will need to be documented in any case - also for the label position.

geofjamg:
Indeed that would be better to be relative to center. That would be a breaking change to change this behaviour but maybe we should consider it?

floriand-e2r:
I do agree. For the node decorators, this can be done in the current PR. For the labels, I can create a breaking change PR.

PR #146, opened 2020-07-23
Replace node

floriand-e2r:

No

Feature

To replace a node by another node, the node needs to be added to the graph and then substituted

A node can be replaced by another node directly by using `Graph.replace` method, which calls `Graph.add` and then `Graph.substitute`

**Other information**:

### DIFF
--- a/single-line-diagram-core/pom.xml
+++ b/single-line-diagram-core/pom.xml
@@ -118,6 +118,11 @@
             <artifactId>powsybl-config-test</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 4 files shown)


floriand-e2r:
The replace method should maybe test if the node is already in the graph... and throw an exception if it is, or deal with a strange replacement (removing then substituting)?

geofjamg:
> The replace method should maybe test if the node is already in the graph... and throw an exception if it is, or deal with a strange replacement (removing then substituting)?

Indeed we need to throw an exception if new node already exists

Issue #182, opened 2020-10-20
Too large blank space between voltage levels


floriand-e2r:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
Substation diagram have a too large width: the blank space between voltage levels is too large; its size corresponds to the size of the first voltage level on its left.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Take the test network in `TestCase11SubstationGraph` (iidm package) and generate the corresponding svg with horizontal substation layout:

        LayoutParameters layoutParameters = new LayoutParameters();
        SubstationGraph g = graphBuilder.buildSubstationGraph(substation.getId(), false);
        new HorizontalSubstationLayoutFactory().create(g, new PositionVoltageLevelLayoutFactory()).run(layoutParameters);
        toSVG(g, "/TestCase11SubstationGraphH.svg", layoutParameters,
            new DefaultDiagramLabelProvider(network, componentLibrary, layoutParameters),
            new DefaultDiagramStyleProvider());

![tooWide](https://user-images.githubusercontent.com/66690739/96621580-130a2f00-1309-11eb-8299-fb0edda13169.png)

* **What is the expected behavior?**
Not such a large blank space between voltage levels

* **What is the motivation / use case for changing the behavior?**
More compact substation diagram.

floriand-e2r:
In the example above, the number of cells can be divided by 2 in method `HorizontalSubstationLayout.calculateCoordVoltageLevel()`. It seems that the cells numbering is increased 2 by 2, as there is one cell above one cell below? But is it always the case?

floriand-e2r:
In the example above, the number of cells can be divided by 2 in method HorizontalSubstationLayout.calculateCoordVoltageLevel(). It seems that the cells numbering is increased 2 by 2, as the maxH corresponds to twice the number of cells.

BenoitJeanson:
Hi,
Sorry for that. it's something I changed a few PR ago. Actually, the elementary step within a voltageLevel Graph is half a cell width. This change was necessary to keep working with integer for (h,v) positions and calculate the intern cell (x,y) coordinates (that are shifted by half a cell width) naturally with no trick as it was neceessary before, when calculating (x,y) coordinates.
I didn't see the impact on substation diagrams.

PR #184, opened 2020-10-21
Extraspace between voltage levels in substation diagrams

floriand-e2r:
PR #180 needs to be merge before this PR.

Yes, #182


FranckLecuyer (commented):
Maybe, the same space reduction could be done for vertical substation
But, if there are many lines or transformers between the different voltage levels, risk is
to have them crossing the voltage levels ....

floriand-e2r:
> But, if there are many lines or transformers between the different voltage levels, risk is
> to have them crossing the voltage levels ....

If I keep the same space as before @BenoitJeanson changes it should be ok I think! But I just realized that I made a mistake about that space in between, which is now divided by 2, I'll correct it right away.

> Maybe, the same space reduction could be done for vertical layout

I should have written it, I already checked for vertical layout: the same problem does not occur as the elementary height (for buses) is still the same as before.

floriand-e2r:
To illustrate here's the svg rendering of TestCase11SubstationGraph substation diagram. The in between space corresponds to what it used to be, that is twice the cell width. That might not be enough in some cases though.
substationInBetweenSpace

FranckLecuyer (approved):

Issue #179, opened 2020-10-20
No space for arrows


floriand-e2r:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
No extra space is left for displaying arrows: they are displayed between feeder and first node/equipment encountered, and often overlap/touch those if there is not enough place.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Display a voltage level with a feeder linked to a bus bar section with two consecutive switches or more, for instance [the example in README.md](https://github.com/powsybl/powsybl-single-line-diagram/raw/master/.github/diagram-demo.svg), where some arrows are below switches, or the following:
![arrowsTooClose3](https://user-images.githubusercontent.com/66690739/96572228-dfa9af00-12cc-11eb-90f3-23c1c4be5f8b.png) ![arrowsTooClose2](https://user-images.githubusercontent.com/66690739/96572229-e0424580-12cc-11eb-8db3-855b85861059.png) ![arrowsTooClose](https://user-images.githubusercontent.com/66690739/96572231-e0dadc00-12cc-11eb-8b18-4cc1a078c7a5.png)

* **What is the expected behavior?**
Some space between the arrows and the switch

* **What is the motivation / use case for changing the behavior?**
Better looking diagrams

floriand-e2r:
In fact there's already code in `PositionVoltageLevelLayout` to free some space for arrows, but that code only shifts the feeders 20 pixels away for that (even if not needed). I can adapt the existing code for now, and maybe in the future we add a fictitious node to avoid possible side effects. What do you think @BenoitJeanson?

BenoitJeanson:
I suggest to add a parameter in LayoutParameters like spaceForFeederArrows. Then, when we want to draw arrows, the calculation of the ExternCell should use (externCellHeight-minSpaceForFeederArrows) space for the nodes between the last node of the LegBlock (the one that is not stuck to the BusNode) and last node before Feeder… and yes sometime an extra ficticious node will be required.
By doing this, a side effect will be that all the arrows will all be aligned: but actually this should be pretty, and readable!

PR #185, opened 2020-10-23
Space for feeder arrows

floriand-e2r:

Yes, #179


floriand-e2r:
A short explanation of what is done:
- Like suggested in the issue, I added a minSpaceForFeederArrows parameter in LayoutParameters, default equal to 50 (5 times the default arrow size)
- I couldn't use LegBlock to deal with Feeders like suggested: indeed, Feeders are not in LegBlocks as it is so far only for blocks connected to the bus. So, I've dealt with Feeders by not counting them in the V-span calculation of BodyPrimaryBlock and setting the Y coordinate of the feeder node from the root node and not in `BodyPrimaryBlock.coordVerticalCase()`.
- As arrowDistance parameter might not be compatible with arrow size and minSpaceForFeederArrows parameter, the two arrows are centered in the available space when inserting them in the svg if there is less than (2 * arrowDistance + 2 * arrow height)

At the end I think that what I've done with this feeder special care inside BodyPrimaryBlock is similar to what is done with LegPrimaryBlock. Maybe it would be neater to include feeders in LegPrimaryBlock, or in a new Block type, to better fit in the model. What do you think @BenoitJeanson?

floriand-e2r:
Two examples to see before / after
arrow-before_2 arrow-after_2 
arrows-before arrows-after

BenoitJeanson:
The code looks very good: that's cool that you refactor the adaptCellHeight in the meantime. The result is pretty.

floriand-e2r:
Some before/after examples (from test files):
shunt
shunt_aligned
3WT_alignment

floriand-e2r:
I checked the results in all test reference files which I needed to update, but in 3 test reference files there are some changes I don't completely understand. They are in the force-layout package which is using ForceAtlas to display substation diagrams. It seems ok to my understanding but I don't know if I'm right... could you tell me if you think it's ok @FranckLecuyer? as you were the one coding that part. I put here the svg files corresponding to the old json reference vs the ones corresponding to the new json reference:
- ForceLayout Compaction Horizontal
forceLayoutHorizontal
- ForceLayout Compaction Vertical
forceLayoutVertical
- ForceLayout Compaction None
forceLayoutNone

geofjamg:
> I checked the results in all test reference files which I needed to update, but in 3 test reference files there are some changes I don't completely understand. They are in the force-layout package which is using ForceAtlas to display substation diagrams. It seems ok to my understanding but I don't know if I'm right... could you tell me if you think it's ok @FranckLecuyer? as you were the one coding that part. I put here the svg files corresponding to the old json reference vs the ones corresponding to the new json reference:
> 
>     * ForceLayout Compaction Horizontal
>       forceLayoutHorizontal
> 
>     * ForceLayout Compaction Vertical
>       forceLayoutVertical
> 
>     * ForceLayout Compaction None
>       forceLayoutNone

Result does not seem to be worst that with previous code just different, but it is not clear to me why by just changing some margins in the diagram, we end up with voltage level in a totally different order...
@FranckLecuyer any idea?

FranckLecuyer:
> > I checked the results in all test reference files which I needed to update, but in 3 test reference files there are some changes I don't completely understand. They are in the force-layout package which is using ForceAtlas to display substation diagrams. It seems ok to my understanding but I don't know if I'm right... could you tell me if you think it's ok @FranckLecuyer? as you were the one coding that part. I put here the svg files corresponding to the old json reference vs the ones corresponding to the new json reference:
> > ```
> > * ForceLayout Compaction Horizontal
> >   forceLayoutHorizontal
> > 
> > * ForceLayout Compaction Vertical
> >   forceLayoutVertical
> > 
> > * ForceLayout Compaction None
> >   forceLayoutNone
> > ```
> 
> Result does not seem to be worst that with previous code just different, but it is not clear to me why by just changing some margins in the diagram, we end up with voltage level in a totally different order...
> @FranckLecuyer any idea?

In the ForceLayout, the voltage levels are initially added with random positions to the undirected graph used by the ForceAtlas algorithm (see ForceSubstationLayout.java, 'run' method).
So, the voltage levels are not always in the same order in the final svg.

FranckLecuyer (approved):

BenoitJeanson (approved):

PR #187, opened 2020-11-04
Write svg in json tests for debug

floriand-e2r:

No.

Feature

One cannot easily generate all svg files related to json reference test files (which sometimes correspond to not-so-good diagrams) to see whether the reference is ok, and when changing the reference file whether that change is not unexpectedly changing the diagram rendering.

If writeFile parameter is true in AbstractTestCase, all json tests inheriting from `AbstractTestCase` will generate the corresponding svg file.

No

**Other information**:
Note that this had already been done by @BenoitJeanson for tests extending `AbstractTestCaseRaw`, this PR simply extends this behaviour to all tests extending `AbstractTestCase`

### DIFF
--- a/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase1.java
+++ b/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase1.java
@@ -37,6 +37,11 @@ import static org.junit.Assert.assertEquals;
  */
 public class TestCase1 extends AbstractTestCaseIidm {
 
+    @Override
+    protected LayoutParameters getLayoutParameters() {
+        return createDefaultLayoutParameters();
+    }
+
     @Before
     public void setUp() {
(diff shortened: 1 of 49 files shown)


single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase1.java
```diff
@@ -61,7 +61,7 @@ public void test() {
         new BlockOrganizer().organize(g);
 
         // calculate coordinates
-        LayoutParameters layoutParameters = new LayoutParameters()
+        layoutParameters = new LayoutParameters()
```
geofjamg, line 64: Maybe not very important but it is probably a little bit weak as a design. Maybe just have an `abstract LayoutParameters getLayoutParameter()` in the base class?

geofjamg (approved):

PR #203, opened 2020-12-02
Default resources component library

floriand-e2r:

No

Feature

Creating a new ResourcesComponentLibrary requires to know in which directory the default components are.

Creating a new ResourcesComponentLibrary can be done without specifying the directory to get the default library.

No


geofjamg (commented):
I don't known if it is a good idea to hard code Convergence style component as default one?

geofjamg:
Instead of doing this, I propose to just create a `ConvergenceComponentLibrary `
```java
class ConvergenceComponentLibrary extends ResourceComponentLibrary {
     ConvergenceComponentLibrary() {
          super("/Convergence");
     }
}
```

geofjamg (changes requested):
For me we don't necessarily need to have a default library and in any case this should not be Convergence one.

PR #159, opened 2020-09-03
add fictitious nodes when connecting a feeder to a busbar in bus brea…

JacquesBorsenberger:
…ker mode

Signed-off-by: jacqbors <jacques.borsenberger@rte-france.com>

**Please check if the PR fulfills these requirements** *(please use `'[x]'` to check the checkboxes, or submit the PR and then click the checkboxes)*
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem ?** *If so, link to this issue using `'#XXX'` and skip the rest*

**What kind of change does this PR introduce?** *(Bug fix, feature, docs update, ...)*

**What is the current behavior?** *(You can also link to an open issue here)*

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?** *If yes, check the following:*
- [ ] The *Breaking Change* or *Deprecated* label has been added
- [ ] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*

**Other information**:

(if any of the questions/checkboxes don't apply, please delete them entirely)

PR #223, opened 2021-02-15
Fix for BusBreakerConnection creation with a switch connected on a empty bus

SlimaneAmar:
- [ x] The commit message follows our guidelines

No

Bug fix

BusBreakerConnection not created with a switch connected on a empty bus

BusBreakerConnection  created with a switch connected on a empty bus

No

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Graph.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Graph.java
@@ -488,7 +488,7 @@ public final class Graph {
     public void substituteFictitiousNodesMirroringBusNodes() {
         getNodeBuses().forEach(busNode -> {
             List<Node> adjs = busNode.getAdjacentNodes();
-            if (adjs.size() == 1 && adjs.get(0).getType() == Node.NodeType.FICTITIOUS) {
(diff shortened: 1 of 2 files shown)


floriand-e2r:
@BenoitJeanson as I don't really get the purpose of removing the fictitious nodes mirroring Bus Nodes, I'm asking for your expertise! if the fictitious nodes mirroring Bus Nodes is a BusBreakerConnection we're losing that connection, as it is removed and subsequently replaced by a fictitious disconnector and fictitious node. As in the changed test case whose result before/after the change is displayed below:
mirroring_before mirroring-after

BenoitJeanson:
This is an old story. I hope I am right. But I remember that in one European  file, each bus (I don't remember if it was of all the substations)  was connected to only one single FictitiousNode which all the cells were connected to.
This ficticiousNode was really just a mirror of the whole busbar. With only externCells, the cellDetector would identify one single megacell per section.

floriand-e2r:
Thx Benoît, I'll try to find that European file for reference! Adding a unit test similar to that file would help in any case, I'll try to do that; it should be in bus breaker view to have both busbreaker connections and that problem.

floriand-e2r:
It seems that 
- the busbreaker connection component introduced in PR #159 
- the fictitious disconnector created if a cell is directly connected to a bus

are two solutions for a same problem, and they collide here. I'll look into how we can combine them.

flo-dup:
BusBreakerConnection was in conflict with fictitious disconnectors, and was therefore removed in PR #241. No need to fix this anymore.

PR #240, opened 2021-05-11
use ServiceLoader.load with the current classloader instead of the context classloader

jonenst:
The context classloader can be the wrong one (and ServiceLoader.load returns an empty list)
when you are calling ServiceLoader.load in the ForkJoinPool.commonPool() (or any other thread pool)
and you have your classes in another classloader than the AppClassLoader (=system classloader).
This is the case when using a springboot fat jar, or when deploying the code in a war in an application
server, or when using mvn exec:java

Signed-off-by: Jon Harper <jon.harper87@gmail.com>

https://github.com/powsybl/powsybl-core/pull/1727


flo-dup (approved):

flo-dup (changes requested):
Don't we also need to change the `ServiceLoaderCache` used in `NetworkGraphBuilder`? Indeed, it may use the wrong context classloader (in `ServiceLoaderCache::getServices`)

jonenst:
@flo-dup  ServiceLoaderCache was modified in core https://github.com/powsybl/powsybl-core/pull/1727 so it should be fixed when we upgrade

flo-dup (approved):

PR #241, opened 2021-05-17
Bus breaker connections refactor

flo-dup:

No

Bug fix

Bus breaker connections have been introduced without considering the existing fictitious disconnectors, which have exactly the same purpose, leading to conflicts and numerous `if (node instanceof BusBreakerConnection) ` patches in the code.

A new fictitious node class is introduced, for both bus breaker and node breaker diagrams, called `BusConnection`. The cell detection and svg drawing is adapted to this new component.

No

**Other information**:
I used a new fictitious node class instead of using the fictitious disconnector for bus breaker to differentiate it, even if the latter was simplifying the cell detection. Indeed, it is more a node than a (operable) disconnector.


flo-dup:
@BenoitJeanson I had a look at what you suggested (that is, adding a new `NodeType` corresponding to the new nodes type `BusConnection`), but I decided to leave it like this as I think it's better to merge this PR and then reconsider the node types in another PR. Indeed, it's not that obvious to add a bus connection type, as these new bus connections are fictitious, so all the fictitious tests should be changed then. And I think we should also reconsider the shunt node type, so that could be done at the same time.

PR #243, opened 2021-05-26
Force layout jgrapht

### LINKED ISSUE #236: License dependency problem
License dependency problem

org.gephi used in force layout has a dependency to colt 1.2.0 library and to concurrent 1.3.4 library trough it, and:
- colt library has a license which forbids military application
- some classes of concurrent have Sun Microsystems copyright (they might not be used by colt library though) 

PowSyBl should not use any copyright library. Besides, for the sake of simplicity, we expect to only use licenses with no limitation.

  - PowSyBl Version: 4.1.0 & sld 2.1.0

fkotd:

Yes, fixes #236

Algorithm change

The ForceSubstationLayout class use gephi’s algorithm to compute force layout.

The ForceSubstationLayout class use a new algorithm from the Springy JavaScript library.


pom.xml
```diff
@@ -167,4 +167,12 @@
         </dependencies>
     </dependencyManagement>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.jgrapht</groupId>
+            <artifactId>jgrapht-core</artifactId>
+            <version>1.5.1</version>
```
flo-dup, line 174: The version is already in the pom parent (powsybl-core). So you can remove the `version` tag

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+        LOGGER.info("Number of steps: {}", iterationCounter);
+    }
+
+    private void applyCoulombsLaw(Graph<Point, Spring> graph) {
+        Set<Point> nodes = graph.vertexSet();
+        for (Point node : nodes) {
+            for (Point otherNode : nodes) {
+                if (!node.equals(otherNode)) {
+                    Vector distance = node.getPosition().subtract(otherNode.getPosition());
+                    double magnitude = distance.magnitude() + 0.1; // avoid massive forces at small distances (and divide by zero)
```
flo-dup, line 58: comment is misleading: it doesn't avoid to divide by zero, because of normalizing below

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+    private double totalEnergy(Graph<Point, Spring> graph) {
+        Set<Point> nodes = graph.vertexSet();
+        double energy = 0.0;
+
+        for (Point node : nodes) {
+            double speed = node.getVelocity().magnitude();
+            energy += 0.5 * node.getMass() * speed * speed;
+        }
+
+        return energy;
```
flo-dup, line 120: you can calculate the total energy with a one-line lambda instead!

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -8,13 +8,8 @@
 
 import com.powsybl.sld.layout.*;
```
flo-dup, line 9: I think the whole file has to be moved out to `single-line-diagram-core` artifact, together with its factory. Indeed the force-layout algorithm could be of interest for other repos which are not related to single-line-diagram.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -77,53 +72,59 @@ public ForceSubstationLayout(SubstationGraph substationGraph,
...
-            n.setPosition(random.nextFloat() * 1000, random.nextFloat() * 1000);
-            undirectedGraph.addNode(n);
+    // TODO: move this function in the SubstationGraph class? It implies that the Substation class see the Point and Edge class which is a bit weird?
+    private Graph<Point, Spring> toJgrapht(SubstationGraph graph) {
+        Graph<Point, Spring> pseudograph = new Pseudograph<>(Spring.class);
+
+        // Create nodes
+        for (VoltageLevelGraph voltageLevelGraph : graph.getNodes()) {
+            String id = voltageLevelGraph.getVoltageLevelInfos().getId();
+            Point point = new Point(id, random.nextDouble() * 1000, random.nextDouble() * 1000);
```
flo-dup, line 82: about the use of random: we need to have the same result from one launch to the other, so we should be able to set the random seed in tests. I'm wondering if it would work as fast (in number of iterations) if we just increase by one the coordinates of each node instead of setting random coordinates?

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -77,53 +72,59 @@ public ForceSubstationLayout(SubstationGraph substationGraph,
...
-    @Override
-    public void run(LayoutParameters layoutParameters) {
-        // Creating the graph model for the ForceLayout algorithm
-        GraphModel graphModel = new GraphModelImpl();
-        UndirectedGraph undirectedGraph = graphModel.getUndirectedGraph();
-        for (VoltageLevelGraph voltageLevelGraph : getGraph().getNodes()) {
-            NodeImpl n = new NodeImpl(voltageLevelGraph.getVoltageLevelInfos().getId());
-            n.setPosition(random.nextFloat() * 1000, random.nextFloat() * 1000);
-            undirectedGraph.addNode(n);
+    // TODO: move this function in the SubstationGraph class? It implies that the Substation class see the Point and Edge class which is a bit weird?
```
flo-dup, line 75: If we make it generic we can more or less move it to `SubstationGraph` class

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/Spring.java
```diff
@@ -0,0 +1,33 @@
+package com.powsybl.sld.force.layout;
+
+import org.jgrapht.graph.DefaultEdge;
+
+public class Spring extends DefaultEdge {
```
flo-dup, line 5: As said earlier, I'm not sure we really need this class. The length and stiffness parameters could be directly set in the force layout algorithm, as they're here the same for all edges. Even if it could be interesting to regroup some nodes together by setting distrinct stiffness and lengths to edges according to some parameters.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/Vector.java
```diff
@@ -0,0 +1,64 @@
+package com.powsybl.sld.force.layout;
+
+import java.util.Objects;
+
+public class Vector {
```
flo-dup, line 5: The name of this class is a bit misleading, as vectors is usually something else! but I'd keep it, as in physic we're always speaking of velocity/acceleration/position *vectors*

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -77,53 +72,59 @@ public ForceSubstationLayout(SubstationGraph substationGraph,
...
+    }
+
+    private Point getPointAtIndex(Graph<Point, Spring> graph, List<Node> adjacentNodes, int index)  {
+        String adjacentNodeId1 = adjacentNodes.get(index).getGraph().getVoltageLevelInfos().getId();
+        return (Point) graph.vertexSet().stream().filter(p -> p.getId().equals(adjacentNodeId1)).findAny().get();
+    }
+
+    @Override
+    public void run(LayoutParameters layoutParameters) {
+        // TODO: rename point and string?
```
flo-dup, line 110: As said above, to deal with a generic graph, we can get rid of `Spring` I think. `Point` I'm not so sure. In fact for each node of the graph, we need to be able to get and set its position, velocity and acceleration. That could be through the `Point` class, I'm ok with that name.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -77,53 +72,59 @@ public ForceSubstationLayout(SubstationGraph substationGraph,
...
-        for (int i = 0; i < maxSteps && forceAtlas2.canAlgo(); i++) {
-            forceAtlas2.goAlgo();
-        }
-        forceAtlas2.endAlgo();
+        return pseudograph;
+    }
+
+    private Point getPointAtIndex(Graph<Point, Spring> graph, List<Node> adjacentNodes, int index)  {
+        String adjacentNodeId1 = adjacentNodes.get(index).getGraph().getVoltageLevelInfos().getId();
+        return (Point) graph.vertexSet().stream().filter(p -> p.getId().equals(adjacentNodeId1)).findAny().get();
```
flo-dup, line 105: Beware, you might be filtering out a big vertex set here, as this can be called many many times. I think we can do better here, but let's see that later, once we use generic graphs.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceSubstationLayout.java
```diff
@@ -77,53 +72,59 @@ public ForceSubstationLayout(SubstationGraph substationGraph,
...
+
+            Point point1 = getPointAtIndex(pseudograph, adjacentNodes, 0);
+            Point point2 = getPointAtIndex(pseudograph, adjacentNodes, 1);
+            pseudograph.addEdge(point1, point2);
 
             if (adjacentNodes.size() == 3) {
-                NodeImpl node3 = (NodeImpl) undirectedGraph.getNode(adjacentNodes.get(2).getGraph().getVoltageLevelInfos().getId());
-                undirectedGraph.addEdge(new EdgeImpl(adjacentEdges.get(1).toString() + "_" + adjacentEdges.get(2).toString(), node2, node3, 0, 1, false));
+                Point point3 = getPointAtIndex(pseudograph, adjacentNodes, 2);
+                pseudograph.addEdge(point2, point3);
```
flo-dup, line 96: I think you forgot to add an edge between `point1` and `point3`

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    private void updateVelocity(Graph<Point, Spring> graph, double deltaTime) {
+        for (Point node : graph.vertexSet()) {
+            Vector velocity = node.getVelocity().add(node.getAcceleration().multiply(deltaTime)).multiply(this.damping);
+            node.setVelocity(velocity);
+
+            if (node.getVelocity().magnitude() > this.maxSpeed) {
+                System.out.println("velocity is superior to max max speed");
```
flo-dup, line 95: Use logger instead

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+            iterationCounter++;
+            if (this.totalEnergy(graph) <= MIN_ENERGY_THRESHOLD) {
+                isStopped = true;
+            }
+            if (iterationCounter >= maxSteps) {
+                isStopped = true;
+            }
+        }
+
+        LOGGER.info("Number of steps: {}", iterationCounter);
```
flo-dup, line 49: It would be interesting to display elapsed time.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+    private final double maxSpeed;
+
+    public ForceLayout() {
+        repulsion = DEFAULT_REPULSION;
+        damping = DEFAULT_DAMPING;
+        maxSpeed = DEFAULT_MAX_SPEED;
+    }
+
+    // TODO: make graph’s node and edge generics instead of point and spring type?
+    public void execute(Graph<Point, Spring> graph, int maxSteps) {
```
flo-dup, line 28: I think we can put `maxSteps` in the constructor with a well chosen default value.

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,122 @@
...
+
+        while (!isStopped) {
+            this.applyCoulombsLaw(graph);
+            this.applyHookesLaw(graph);
+            this.attractToCenter(graph);
+            this.updateVelocity(graph, deltaTime);
+            this.updatePosition(graph, deltaTime);
+
+            iterationCounter++;
+            if (this.totalEnergy(graph) <= MIN_ENERGY_THRESHOLD) {
```
flo-dup, line 41: Can we avoid unnecessary iterations by putting the threshold a bit higher? Do you think it would be interesting to also put that threshold in the constructor? so that the user can parameterize the convergence criteria.

flo-dup (changes requested):
It needs some rework I think, as you suggested, but this is a great first version!

single-line-diagram-force-layout/src/main/java/com/powsybl/sld/force/layout/ForceLayout.java
```diff
@@ -0,0 +1,256 @@
...
+
+    public ForceLayout<V, E> setInitialisationSeed(long seed) {
+        random.setSeed(seed);
+        return this;
+    }
+
+    // TODO: implement other initialisations methods
+    public ForceLayout<V, E> initializePoints() {
+        for (V vertex : this.graph.vertexSet()) {
+            // TODO: not sure vertex is unique anymore if the user override the equals method
```
flo-dup, line 85: As it is a set, you can be sure vertex is unique, as the set uses the equals method to contiain each element only once

flo-dup (approved):

PR #253, opened 2021-07-05
Generating the real component type in node metadata for 2WT and 3WT transformer feeder leg node

FranckLecuyer:

Bug fix

The componentType in node metadata for 2WT or 3WT feeder leg node is set to LINE

The componentType in node metadata for 2WT or 3WT feeder leg node is now set to TWO_WINDINGS_TRANSFORMER ot THREE_WINDINGS_TRANSFORMER

**Other information**:


single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -499,9 +499,19 @@ protected void setMetadata(GraphMetadata metadata, Node node, String nodeId, Vol
             }
         }
 
+        // to generate the real component type in node metadata for 2WT and 3WT transformers
+        // (actually, the component type is set to 'LINE' for 3WT feeder leg nodes and for 2WT feeder leg nodes in
+        //  substation diagram, for internal reasons)
```
flo-dup, line 504: I checked it, and unless I forgot something there's no internal reason (anymore?) to set the component type to `LINE` for the leg nodes.

flo-dup (changes requested):
I don't think the _real_ component type for the transformers leg nodes is `TWO_WINDINGS_TRANSFORMER` or `THREE_WINDINGS_TRANSFORMER`. Besides it's strange to have two nodes (the feeder node + the component node) with that type for the same transformer in the metadata. What do you think of introducing a new component type, like `WINDINGS_TRANSFORMER_LEG`?

flo-dup (approved):

PR #283, opened 2021-09-24
Fix color trick

### LINKED ISSUE #250: Fix color trick
Bug 

Fill color according to voltage levels is not supported, hence a trick is used by having a stroke thick enough to fill the wanted shape:
   - see in `DefaultSVGWriter::drawNodeInfos` the radius set to half the requested value and the stroke width
   - see `bus_connection.svg` resource file with a radius of 2 instead of 4 and a stroke-width of 4 defined in the `components.css` file

Fill color according to voltage levels supported 

Avoid using a trick to get the shape filled

* **Other information**
CSS variable value should probably be used to avoid duplication of fill and stroke colors. Each voltage CSS class would define the same variable (for instance `.sld-vl180to300 {--sld-vl-color: #228b22}`) and then the components style classes decide what to do with that color (for instance `.sld-wire {stroke: var(--sld-vl-color)}` or `.sld-bus-connection {fill: var(--sld-vl-color)}`). This would also allow to get rid of both `DefaultDiagramStyleProvider::isConstantColor` methods

tadam50:

Fixes #250

Bug fix

Using stroke-width to fill circle node information

Using fill parameter with css variable to fill circle node information

**Does this PR introduce a breaking change or deprecate an API?** yes


single-line-diagram-core/src/main/resources/topologicalBaseVoltages.css
```diff
@@ -1,72 +1,71 @@
...
+.sld-vl0to30-0 {--sld-vl-color: #AAAE27; stroke:var(--sld-vl-color)}
+.sld-vl0to30-1 {--sld-vl-color: #195B0F; stroke:var(--sld-vl-color)}
+.sld-vl0to30-2 {--sld-vl-color: #EABABE; stroke:var(--sld-vl-color)}
+.sld-vl0to30-3 {--sld-vl-color: #2D6C13; stroke:var(--sld-vl-color)}
+.sld-vl0to30-4 {--sld-vl-color: #DDA193; stroke:var(--sld-vl-color)}
+.sld-vl0to30-5 {--sld-vl-color: #477D17; stroke:var(--sld-vl-color)}
+.sld-vl0to30-6 {--sld-vl-color: #CE9A6E; stroke:var(--sld-vl-color)}
+.sld-vl0to30-7 {--sld-vl-color: #648D1C; stroke:var(--sld-vl-color)}
+.sld-vl0to30-8 {--sld-vl-color: #BEA04A; stroke:var(--sld-vl-color)}
+.sld-vl0to30-9 {--sld-vl-color: #869D22; stroke:var(--sld-vl-color)}
```
flo-dup, line 71: Like I said in baseVoltages.css I'd prefer to put the stroke in the `components.css` file

flo-dup (changes requested):

single-line-diagram-core/src/main/resources/ConvergenceLibrary/components.json
```diff
@@ -432,32 +432,6 @@
       "styleClass" : null
     } ],
     "styleClass" : "sld-flash"
-  }, {
-    "type" : "BUSBAR_SECTION",
```
flo-dup, line 436: woops good catch! missed it in previous PR.

single-line-diagram-core/src/main/resources/tautologies.css
```diff
@@ -10,3 +9,20 @@ svg {fill: none; stroke: black}
...
+.sld-generator {fill: none}
+.sld-two-wt {fill: none}
+.sld-three-wt {fill: none}
+.sld-capacitor {fill: none}
+.sld-inductor {fill: none}
+.sld-pst {fill: none}
+.sld-svc {fill: none}
+.sld-vsc {fill: none}
+/* Stroke blue & fill: white */
+.sld-breaker {stroke: blue; fill: white}
```
flo-dup, line 28: I think these lines should be in components.css instead. What I meant with this "tautologies" file is that it contains properties that remain true whatever comes next. Which is not the case of the components as they may be filled with a color for a specific component library.

single-line-diagram-core/src/main/resources/baseVoltageConstantColors.css
```diff
@@ -1,5 +1,12 @@
+/* ----------------------------------------------------------------------- */
+/* File: baseVoltageConstantColors.css ----------------------------------- */
+/* Stroke none & fill: black */
+.sld-label {stroke: none; fill: black}
 .sld-node {stroke: none; fill: black}
```
flo-dup, line 5: I think all style related to components should be in components.css. There won't be a lot left then here.

single-line-diagram-core/src/main/resources/defaultConstantColors.css
```diff
@@ -1,9 +1,26 @@
...
+.sld-load {stroke: blue}
+.sld-generator {stroke: blue}
+.sld-two-wt {stroke: blue}
+.sld-three-wt {stroke: blue}
+.sld-winding {stroke: blue}
+.sld-capacitor {stroke: blue}
+.sld-inductor {stroke: blue}
+.sld-pst {stroke: blue}
+.sld-svc {stroke: blue}
+.sld-vsc {stroke: blue}
```
flo-dup, line 26: As said before, I think all style related to components should be in components.css. I think this file can be merged with baseVoltageConstantColors.css, as at the end there won't be anything specific with base voltage there. In fact both can maybe be deleted, as only labels are not components but their style is linked to components (and there's already an entry for labels in components.css). What do you think?

flo-dup (changes requested):

flo-dup (approved):

flo-dup (approved):

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -1321,9 +1321,9 @@ private void drawNodeInfos(ElectricalNodeInfo nodeInfo, double xShift, double yS
         circle.setAttribute("cy", String.valueOf(yShift));
         circle.setAttribute("r", String.valueOf(CIRCLE_RADIUS_NODE_INFOS_SIZE));
         circle.setAttribute("stroke-width", String.valueOf(CIRCLE_RADIUS_NODE_INFOS_SIZE));
-        if (!styles.isEmpty()) {
-            circle.setAttribute(CLASS, String.join(" ", styles));
-        }
+        List<String> classStyles = new ArrayList<>(styles);
+        classStyles.add(NODE_INFOS);
```
flo-dup, line 1325: I think that this `NODE_INFOS` style should be directly added in the `AbstractBaseVoltageDiagramStyleProviders::getBusStyles` & `DefaultDiagramStyleProvider::getBusStyles`

flo-dup (changes requested):

flo-dup (approved):

PR #297, opened 2021-11-05
Labels and decorators on all nodes (#293)

### LINKED ISSUE #293: Labels and decorators on all nodes
Feature

Labels and decorators cannot be added on multiTermNodes and fictitious nodes

Labels and decorators can be added on multiTermNodes and fictitious nodes

MultitermNodes: Transformers between voltage levels might need a label
Ficitious nodes: no real use case, but no reason to forbid it?

tadam50:

Fixes #293 

Feature

Labels and decorators cannot be added on multiTermNodes and fictitious nodes

Labels and decorators can be added on multiTermNodes and fictitious nodes

**Other information**:

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
@@ -12,6 +12,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -116,13 +117,14 @@ public class Node implements BaseNode {
         return equipmentId;
     }
 
-    public String getLabel() {
+    public Optional<String> getLabel() {
+        String result = null;
         if (label != null) {
-            return label;
+            result = label;
         } else if (graph != null) {
-            return graph.isUseName() ? name : equipmentId;
+            result = graph.isUseName() ? name : equipmentId;
         }
-        return null;
+        return Optional.ofNullable(result);
     }
 
     public void setLabel(String label) {
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
@@ -122,7 +122,6 @@ public class Node implements BaseNode {
         } else if (graph != null) {
             return graph.isUseName() ? name : equipmentId;
         }
-        // FIXME : to be confirmed
         return null;
     }
 
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
@@ -119,9 +119,11 @@ public class Node implements BaseNode {
     public String getLabel() {
         if (label != null) {
             return label;
-        } else {
+        } else if (graph != null) {
             return graph.isUseName() ? name : equipmentId;
         }
+        // FIXME : to be confirmed
+        return null;
     }
 
     public void setLabel(String label) {
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramLabelProvider.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramLabelProvider.java
@@ -115,11 +115,11 @@ public class DefaultDiagramLabelProvider implements DiagramLabelProvider {
 
         List<NodeLabel> nodeLabels = new ArrayList<>();
         if (node instanceof FeederNode) {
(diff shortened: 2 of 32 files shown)


single-line-diagram-core/src/main/java/com/powsybl/sld/model/Node.java
```diff
@@ -119,9 +119,11 @@ public String getEquipmentId() {
     public String getLabel() {
         if (label != null) {
             return label;
-        } else {
+        } else if (graph != null) {
             return graph.isUseName() ? name : equipmentId;
         }
+        // FIXME : to be confirmed
+        return null;
```
flo-dup, line 126: I confirm! the line above avoids a NPE

single-line-diagram-core/src/test/resources/NodeDecoratorsBranchStatusNodeBreaker.svg
```diff
@@ -812,6 +812,9 @@ polyline {fill: none}
             <circle cx="5" cy="15" r="5"/>
             <circle cx="11" cy="15" r="5"/>
             <circle cx="8" cy="19" r="5"/>
+            <g class="sld-flash">
+                <path d="M 5.5,14.5 8.5,6.36 4.32,5.83 4.2,0.3 1.3,8.47 5.4,8.99 Z" transform="translate(3.0,-0.5)"/>
+            </g>
```
flo-dup, line 817: This flash should not be added... It's due to using Middle3WTNode for two different usages: it is either a node among `voltageLevelGraph.getNodes()` or a multiTermNode of a `SubstationGraph` or `VoltageLevelGraph`.  We should display a flash only in the first case, and to know you're in this case you need to test if `middle3wtNode.getGraph() != null` (in `DefaultDiagramLabelProvider::getNodeDecorators`).

flo-dup (changes requested):

single-line-diagram-core/src/test/java/com/powsybl/sld/AbstractTestCase.java
```diff
@@ -76,7 +76,7 @@ String getName() {
 
     private void writeToFileInHomeDir(String filename, StringWriter content) {
         File homeFolder = new File(System.getProperty("user.home"));
-        File file = new File(homeFolder, filename);
+        File file = new File(homeFolder + File.separator + ".powsybl", filename);
```
flo-dup, line 79: Why not? It sure is annoying to get all those files directly in your home folder. But you'd better do it in another PR.

flo-dup (changes requested):
You replaced many (all?) json unit tests with SVG tests, could you revert that? I mentioned adding a single test with `RawDiagramLabelProvider`, but one is enough. Speaking of that, some people would like more SVG tests instead of json tests, but anyway that's not the goal of this PR.

single-line-diagram-core/src/test/resources/InternalBranchesBusBreaker.svg
```diff
@@ -299,6 +299,7 @@ polyline {fill: none}
                 <circle cx="5" cy="15" r="5" transform="rotate(180.0,8.0,12.0)"/>
                 <circle cx="11" cy="15" r="5" transform="rotate(180.0,8.0,12.0)"/>
                 <circle cx="8" cy="19" r="5" transform="rotate(180.0,8.0,12.0)"/>
+                <text class="sld-label" id="T3_12_T3_12_NW_LABEL" x="-5.0" y="-5.0">T3_12</text>
```
flo-dup, line 302: In fact adding such a label is the goal of current PR, but I think we don't want that for the default behaviour, therefore the `DefaultDiagramLabelProvider::getNodeLabels` should not add anything for the fictitious nodes

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramLabelProvider.java
```diff
@@ -115,9 +115,11 @@ public DefaultDiagramLabelProvider(Network net, ComponentLibrary componentLibrar
 
         List<NodeLabel> nodeLabels = new ArrayList<>();
         if (node instanceof FeederNode) {
-            nodeLabels.add(new NodeLabel(node.getLabel(), getFeederLabelPosition(node)));
+            node.getLabel().ifPresent(label -> nodeLabels.add(new NodeLabel(label, getFeederLabelPosition(node), null)));
         } else if (node instanceof BusNode) {
-            nodeLabels.add(new NodeLabel(node.getLabel(), getBusLabelPosition(node)));
+            node.getLabel().ifPresent(label -> nodeLabels.add(new NodeLabel(label, getBusLabelPosition(node), null)));
+        } else if (node instanceof FictitiousNode) {
+            node.getLabel().ifPresent(label -> nodeLabels.add(new NodeLabel(label, getBusLabelPosition(node), null)));
```
flo-dup, line 122: So that's the unwanted default behaviour, you can remove that condition

flo-dup (changes requested):

flo-dup (approved):

PR #309, opened 2021-12-02
Remove specific prefix in base-voltages.yml

flo-dup:

No

Bug fix

There is a specific prefix in base-voltages.yml, leading to conflicts when using network-area-diagram at the same time

The specific prefix has been removed, it is added afterwards in the code.

No


geofjamg (approved):

geofjamg (changes requested):
After thinking again and trying to use it with NAD, in pypowsybl, I am not sure it is the right thing to do. Maybe keeking the prefix + changing file name to sld-base-voltages.yml would be better to avoid having 2 identical files (mais not sure it will remains...) in the classpath.

flo-dup:
> After thinking again and trying to use it with NAD, in pypowsybl, I am not sure it is the right thing to do. Maybe keeking the prefix + changing file name to sld-base-voltages.yml would be better to avoid having 2 identical files (mais not sure it will remains...) in the classpath.

One commit was missing here, the base voltages single-line-diagram file needed to be removed, and a default file added in core (see powsybl/powsybl-core#1919). Adding this commit makes the test fail though - didn't want to point to core snapshot version, let's wait for the RC.

PR #341, opened 2022-01-25
Up to N FeederInfos

### LINKED ISSUE #303: Up to N FeederInfos (arrows)
Feature

Only 2 FeederInfos expected

Add more that 2 FeederInfos

Some users need to add up to 5 arrows (to display 5 different values)

* **Other information**
Maybe everything's already working thanks to #284, except the space reserved to draw the arrows. Indeed, the cell height calculation shoud take into account the number of arrows. It is up to now done by `PositionVoltageLevelLayout::getFeederSpan`, which calls `LayoutParameters::getMinSpaceForFeederArrows`. We could just change that minSpaceForFeederArrows to display more arrows, but I think instead of a `minSpaceForFeederArrows` there should be a `nbMaxFeederInfos` parameter, so that the user only set this instead of number of pixel, which is less user-friendly.
Note that calling the LabelProvider to know how many arrows will be displayed when calculating the cell height is not unfortunately possible, as this provider was meant only for SVG drawing and is therefore not available during layout.

tadam50:

Fixes #303 

Feature

Only 2 FeederInfos expected

Add more that 2 FeederInfos

**Other information**:


single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -785,20 +785,20 @@ protected void transformArrow(List<Point> points, ComponentSize componentSize, d
...
+            // If not enough space to have layoutParameters.getArrowDistance() at both sides of the 2 feeder infos,
+            // we compute the distance between feeder anchor and first feeder info so that the two feeder infos are centered.
+            double distFeederAnchorToFirstFeederInfoCenter =
+                distancePoints >= 2 * layoutParameters.getFeederInfosOuterMargin() + 2 * componentSize.getHeight()
+                    ? layoutParameters.getFeederInfosOuterMargin()
                     : (distancePoints - 2 * componentSize.getHeight()) / 2;
-            double x = pointA.getX() + cosAngle * (distFeederAnchorToFirstArrowCenter + shift);
-            double y = pointA.getY() + sinAngle * (distFeederAnchorToFirstArrowCenter + shift);
+            double x = pointA.getX() + cosAngle * (distFeederAnchorToFirstFeederInfoCenter + shift);
+            double y = pointA.getY() + sinAngle * (distFeederAnchorToFirstFeederInfoCenter + shift);
```
flo-dup, line 796: Could you delete this code? I think there's no use to keep it now that the user is responsible for giving enough space for the feederInfos

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -851,15 +851,15 @@ protected void insertFeederInfos(String prefixId,
             points.add(new Point(feederNode.getDiagramCoordinates()));
         }
 
-        int iArrow = 0;
+        double shiftFeederInfo2 = 0;
```
flo-dup, line 854: why the 2?

flo-dup (changes requested):

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
tadam50: My mistake !

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
tadam50: I've change as following : 
double distFeederAnchorToFirstFeederInfoCenter = layoutParameters.getFeederInfosOuterMargin()

flo-dup (approved):

PR #346, opened 2022-02-08
Adding order and direction to SwitchNode serialization

### LINKED ISSUE #340: Serialized SwitchNode looses "order and "direction" properties
A bug

When serialized to Json, SwitchNodes are loosing their 'order" and "direction" properties.

1. Create a SwitchNode with an order and a direction.
2. Serialize it using "writeJsonContent" method
3. Deserialize it.

Order and direction are now null values.

Order and direction values should remain the same as before serialization.

* **What is the motivation / use case ?**
We need to serialize graphs in order to store them into a database

  - PowSyBl Version: 2.6.0

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

Workaround : we created a CustomSwitchNode class extending SwitchNode, overriding writeJsonContent method.

MarinPinchinatLoth:

Yes, fixes #340 by adding optional order and direction to SwitchNode serialization.


single-line-diagram-core/src/main/java/com/powsybl/sld/model/SwitchNode.java
```diff
@@ -50,5 +50,11 @@ public Node getOtherAdjNode(Node adj) {
     protected void writeJsonContent(JsonGenerator generator, boolean includeCoordinates) throws IOException {
         super.writeJsonContent(generator, includeCoordinates);
         generator.writeStringField("kind", kind.name());
+        if (this.getOrder().isPresent()) {
```
flo-dup, line 53: you can remove the `this.` keyword in the added lines to slightly reduce the code

single-line-diagram-core/src/main/java/com/powsybl/sld/model/SwitchNode.java
```diff
@@ -50,5 +50,11 @@ public Node getOtherAdjNode(Node adj) {
...
+            generator.writeNumberField("order", this.getOrder().get());
```
flo-dup, line 54: You need to store the result of `getOrder` as nothing ensures the second call returns a non-empty optional. And therefore sonar detects it as a bug.

It would be even more readable to use `Optional::ifPresent`, but unfortunately you cannot because of the IOException

flo-dup (changes requested):
Thanks for fixing that bug! One more thing, could you amend your commit to add your signoff? and then force-push (see [DCO details](https://github.com/powsybl/powsybl-single-line-diagram/pull/346/checks?check_run_id=5112087938) for more info if needed)

flo-dup (approved):

PR #354, opened 2022-02-21
Possibility to add indicators on busbarsection

### LINKED ISSUE #311: Possibility to add indicators on busbarsection
Feature

See the attached picture:
image

We would like to add the voltage lack information on the bus bar section if available

  - PowSyBl Version: ...
  - OS Version: ...

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)

(if a question doesn't apply, you can delete it)

tadam50:

Fixes #311 

feature

No indicator provided on busbar section

Be able to add indicator on busbar section

No


flo-dup (commented):
Great! Similarly to the FeederInfo you can now add a BusInfo given by the LabelProvider to draw something in that new cell

tadam50:
Indicators drawn with bus style as following :
TestCase15GraphWithVoltageLackInfo

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +902,47 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
         root.appendChild(g);
     }
 
+    protected void insertBusInfo(String prefixId,
+                                 Point point,
```
flo-dup, line 906: `point` is the shift due to the voltage level coordinates, so can you choose a clearer name? like vlPoint or shift or vlShift?

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +902,47 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
...
+        Optional<BusInfo> busInfo = initProvider.getBusInfo(busNode);
+        busInfo.ifPresent(info -> {
+            drawBusInfo(prefixId, busNode, point, root, busInfo.get(), styleProvider, metadata);
+            addInfoComponentMetadata(metadata, busInfo.get().getComponentType(), false);
+        });
+    }
+
+    private void drawBusInfo(String prefixId,
+                             BusNode busNode,
+                             Point point,
```
flo-dup, line 921: same here

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +902,47 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
...
+                             Element root,
+                             BusInfo busInfo,
+                             DiagramStyleProvider styleProvider,
+                             GraphMetadata metadata) {
+        Element g = root.getOwnerDocument().createElement(GROUP);
+        // Position
+        ComponentSize componentSize = componentLibrary.getSize(busInfo.getComponentType());
+        double shY = componentSize.getHeight() / 2;
+        Point origin = new Point(point.getX(), point.getY() - shY);
```
flo-dup, line 930: here you choose to put the component exactly on the bus bar extremity, but I think we should add a parameter if a user wants to shift it a bit on the bus bar (like shiftX? because I think there's no use to shift it on the Y axis)

single-line-diagram-core/src/main/resources/ConvergenceLibrary/components.json
```diff
@@ -410,5 +410,17 @@
       "styleClass" : "sld-arrow-in"
     } ],
     "styleClass" : "sld-arrow-q"
+  }, {
+    "type" : "LACK_VOLTAGE",
```
flo-dup, line 414: I'm wondering if we should provide this component except in tests, as there's no way to add it to a network with the exisiting GraphBuilders. But that's maybe not the only component in that case in fact!

flo-dup (commented):
Great job! I think we can now go a bit further, and add a cell on the left and on the right, and add an enum parameter on the BusInfo which makes the drawing on the left or right side of the bus.

single-line-diagram-core/src/main/resources/ConvergenceLibrary/components.json
tadam50: How append lack-voltage component to existing ConvergenceLibrary only for tests ?

single-line-diagram-core/src/main/java/com/powsybl/sld/layout/BlockOrganizer.java
```diff
@@ -34,6 +34,8 @@
 
     private final boolean exceptionIfPatternNotHandled;
 
+    private final boolean voltageIndicatorOnBus;
```
flo-dup, line 37: voltageIndicator is only one possible use for busInfo, so we should keep that keyword for the test only

single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LayoutParameters.java
```diff
@@ -78,6 +78,12 @@
 
     private double feederInfosIntraMargin = 10;
 
+    /**
+     * Can be used as horizontal shifting value for busInfo indicator.
+     * Shifting to the left with negative value and the right with positive value.
```
flo-dup, line 83: Just say that it could be negative instead of that sentence (which is not exactly right), I think it's pretty clear as you called it margin how the shift should be done.

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/BusInfo.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private final String userDefinedId;
+
+    private final String topLabel;
+
+    private final String bottomLabel;
+
+    private final Side anchor;
+
+    private final boolean powered;
```
flo-dup, line 37: As said earlier, I think this boolean narrows down the use of BusInfo, which could maybe be used for something else. Therefore it's better in the voltage indicator test to have two components instead of just one.

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/BusInfo.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public BusInfo(String componentType, boolean powered, String topLabel, String bottomLabel) {
```
flo-dup, line 39: Could you add a constructor with only componentType as argument?

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +898,67 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
...
 
+    protected void insertBusInfo(String prefixId,
+                                 Point vlShift,
+                                 double shiftX,
+                                 Element root,
+                                 BusNode busNode,
+                                 GraphMetadata metadata,
+                                 DiagramLabelProvider initProvider,
+                                 DiagramStyleProvider styleProvider) {
+        Optional<BusInfo> busInfo = initProvider.getBusInfo(busNode);
```
flo-dup, line 909: you can remove `busInfo` variable, you don't need it (use `info` in the lambda instead of `busInfo.get()`)

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +898,67 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
...
+    private void drawBusInfo(String prefixId,
+                             BusNode busNode,
+                             Point vlShift,
+                             double shiftX,
+                             Element root,
+                             BusInfo busInfo,
+                             DiagramStyleProvider styleProvider,
+                             GraphMetadata metadata) {
+        Element g = root.getOwnerDocument().createElement(GROUP);
+        // Position
```
flo-dup, line 925: I find it more readable when there's an empty line before the comment, would you agree?

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DiagramStyles.java
```diff
@@ -36,6 +36,8 @@
     public static final String TOP_FEEDER = STYLE_PREFIX + "top-feeder";
     public static final String FRAME_CLASS = STYLE_PREFIX + "frame";
     public static final String NODE_INFOS = STYLE_PREFIX + "node-infos";
+    public static final String POWERED_CLASS = STYLE_PREFIX + "powered";
+    public static final String UNPOWERED_CLASS = STYLE_PREFIX + "unpowered";
```
flo-dup, line 40: Following what I said before, those two classes should be in the components.css of the test, one for each component.

single-line-diagram-core/src/test/resources/VoltageIndicatorLibrary/components.css
```diff
@@ -0,0 +1,7 @@
+/* ----------------------------------------------------------------------- */
+/* File : components.css ------------------------------------------------- */
+.sld-voltage-indicator .sld-label {dominant-baseline: middle}
+.sld-voltage-indicator .sld-unpowered {stroke: none; fill: grey}
+.sld-voltage-indicator .sld-powered {stroke: none; fill: var(--sld-vl-color, black)}
```
flo-dup, line 5: Ok, I got it, keeping only one component makes the things nicer here...
I think a way to go could be to keep one component only, but to add a new `BusVoltageInfo` class in the test which extends `BusInfo` and adds this annoying boolean. Then we need to add in the style provider interface a `Optional<String> getBusInfoStyle(BusInfo)` which you will override in the test, giving the style according to the boolean value.

single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase15GraphWithVoltageIndicator.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testTopological() {
+        runTest(new TopologicalStyleProvider(network), "/TestCase15GraphWithVoltageIndicatorTopological.svg");
+    }
+
+    @Test
+    public void testNominal() {
+        runTest(new NominalVoltageDiagramStyleProvider(network), "/TestCase15GraphWithVoltageIndicatorNominal.svg");
+    }
+
```
flo-dup, line 102: I think you can remove that test, the result is almost the same than previous one, and we don't lose coverage by removing it.

flo-dup (changes requested):

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -422,6 +425,8 @@ protected void drawBuses(String prefixId,
                         true, null));
             }
 
+            insertBusInfo(prefixId, graph.getCoord(), layoutParameters.getBusInfoMargin(), root, busNode, metadata, initProvider, styleProvider);
```
flo-dup, line 428: I think it makes more sense to put the BusInfo directly inside the BusNode Element `g`. Just change `root` by `g` in this line, and move it above `root.appendChild(g)`

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +898,71 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
...
+        ComponentSize size = componentLibrary.getSize(busInfo.getComponentType());
+        double vlShiftX = vlShift.getX() + shiftX;
+        if (busInfo.getAnchor() == Side.RIGHT) {
+            vlShiftX = vlShift.getX() + busNode.getPxWidth() - shiftX - size.getWidth();
+        }
+        Point origin = new Point(vlShiftX, vlShift.getY() - size.getHeight() / 2);
+        transformComponent(busNode, origin, g);
+
+        // Styles
+        List<String> styles = styleProvider.getSvgNodeStyles(busNode, componentLibrary, layoutParameters.isShowInternalNodes());
```
flo-dup, line 936: I saw that sld-busbar-section was added there, which is strange. That's how I came up with the comment that BusInfo should be inside the BusNode svg element. Once it would be inside that element, it will inherit its style, and therefore you can replace this line by `new ArrayList<>()`. And you will need to adjust the shift also!

single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -899,6 +898,71 @@ private void drawFeederInfo(String prefixId, String feederNodeId, List<Point> po
         root.appendChild(g);
     }
 
+    protected void insertBusInfo(String prefixId,
+                                 Point vlShift,
+                                 double shiftX,
```
flo-dup, line 903: You can remove shiftX argument, drawBusInfo can get it directly from the parameters. It makes sense to keep it there if this shift was not the same for all the BusInfo, but I don't think it's useful so far.

flo-dup (changes requested):

single-line-diagram-core/src/main/java/com/powsybl/sld/layout/BlockOrganizer.java
flo-dup: maybe `addCellForBusInfo`, what do you think?

flo-dup (approved):

PR #363, opened 2022-03-07
Adding one cell on previous subsection only at right or left side

tadam50:

No

Feature

When addCellForBusInfo parameter of BlockOrganizer is set to true, one cell is adding on previous subsection only at right and left side

Replace addCellForBusInfo parameter of BlockOrganizer by a map (bus id / side) 
in order to add cell on previous subsection only at right or left side

**Other information**:

### DIFF
--- a/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DiagramLabelProvider.java
+++ b/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DiagramLabelProvider.java
@@ -10,9 +10,11 @@ import com.powsybl.sld.model.BusNode;
 import com.powsybl.sld.model.FeederNode;
 import com.powsybl.sld.model.Node;
 import com.powsybl.sld.model.VoltageLevelGraph;
+import com.powsybl.sld.model.coordinate.Side;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DiagramLabelProvider.java
```diff
@@ -90,4 +92,8 @@ public LabelPosition getPosition() {
     default Optional<BusInfo> getBusInfo(BusNode node) {
         return Optional.empty();
     }
+
+    default Map<String, Side> getBusInfoSides(VoltageLevelGraph graph) {
+        return Collections.emptyMap();
```
flo-dup, line 97: In fact I thought of doing directly like:
`graph.getNodeBuses().forEach(busNode -> getBusInfo(busNode).ifPresent(busInfo -> result.put(busNode.getId(), busInfo.getAnchor())));`

flo-dup (changes requested):

flo-dup:
About the unit tests, could you test some corner-cases? like a subsection with no BusInfo at all. For instance remove the top and bottom in the middle (busInfo for bbs1 and bbs22) in one of the two unit tests.

PR #365, opened 2022-03-08
Add & use getInnerHeight

### LINKED ISSUE #280: Snakelines bug
Bug

Snakelines assume the height of each voltage level is constant

See diagram-demo.svg in README.md

Two snakelines can't have the same ordinate

Avoid overlapping

  - PowSyBl Version: sld 2.4.0 but was already occuring before

tadam50:

Fixes #280 

Bug fix

Snakelines assume the height of each voltage level is constant

Two snakelines can't have the same ordinate


single-line-diagram-core/src/main/java/com/powsybl/sld/layout/VerticalSubstationLayout.java
```diff
@@ -117,8 +117,11 @@ private double getHeightHorizontalSnakeLines(String vlGraphId, BusCell.Direction
             infosNbSnakeLinesH.getNbSnakeLinesTopBottom().put(BusCell.Direction.TOP, currentNbTop);
             infosNbSnakeLinesH.getNbSnakeLinesVerticalBetween().put(graphId, currentNbLeft);
 
+            double yMin = getGraph().getVoltageLevels().stream().mapToDouble(g -> g.getY()).min().orElse(0.0);
+            double yMax = getGraph().getVoltageLevels().stream().mapToDouble(g -> g.getY() + g.getInnerHeight(layoutParam)).max().orElse(0.0);
+
             // Calculate the snakeline as an horizontal layout
-            polyline = calculatePolylineSnakeLineForHorizontalLayout(layoutParam, node1, node2, increment, infosNbSnakeLinesH);
+            polyline = calculatePolylineSnakeLineForHorizontalLayout(layoutParam, node1, node2, increment, infosNbSnakeLinesH, yMin, yMax);
```
flo-dup: Here we are in a substation vertical layout, but within a voltage level (snake lines with two ends in the same voltage level). So you need to calculate yMin and yMax by considering only the current voltage level. Indeed the other voltage levels will be above / below.

flo-dup (changes requested):

flo-dup (approved):

single-line-diagram-core/src/main/java/com/powsybl/sld/layout/HorizontalSubstationLayout.java
```diff
@@ -27,7 +27,9 @@ public HorizontalSubstationLayout(SubstationGraph graph, VoltageLevelLayoutFacto
     @Override
     protected List<Point> calculatePolylineSnakeLine(LayoutParameters layoutParam, Node node1, Node node2,
                                                      boolean increment) {
-        return calculatePolylineSnakeLineForHorizontalLayout(layoutParam, node1, node2, increment, infosNbSnakeLines);
+        double yMin = getGraph().getVoltageLevels().stream().mapToDouble(VoltageLevelGraph::getY).min().orElse(0.0);
+        double yMax = getGraph().getVoltageLevels().stream().mapToDouble(g -> g.getY() + g.getInnerHeight(layoutParam)).max().orElse(0.0);
```
flo-dup, line 31: Just one thing I forgot, to avoid calculating it for each node we could calculate yMin & yMax in AbtractLayout::manageSnakeLines and pass it as arguments.

flo-dup (changes requested):

single-line-diagram-core/src/main/java/com/powsybl/sld/layout/HorizontalSubstationLayout.java
tadam50: > Just one thing I forgot, to avoid calculating it for each node we could calculate yMin & yMax in AbtractLayout::manageSnakeLines and pass it as arguments.

For VerticalLayout case, we need to compute it for each node, right ?

flo-dup: yes indeed! let's forget about this, and I'll create an issue to create yMin / yMax attribute for substation graphs, or more generally for any kind of graphs

Issue #191, opened 2020-11-12
Flip horizontally components in bottom direction cells 


floriand-e2r:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Components can be rotated when being in bottom direction cells but not flipped horizontally

* **What is the expected behavior?**
Components can be easily flipped horizontally if in bottom direction cells

* **What is the motivation / use case for changing the behavior?**
It allows to have a symmetry relative to the busbar even if components have no vertical symmetry.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, spectrum, etc)
Created following this [spectrum question](https://spectrum.chat/powsybl/questions/release-sld-1-6-0-flip-horizontally-a-component~01ba5d4f-12ac-44f1-b018-30ff6367e7d1).

flo-dup:
I suggest following changes:
- add Orientation parameter to each `Node` (serialized or not?)
- replace `allowRotation` in `Component` class by 2  parameters
   - one for horizontal transformation (`LEFT` or `RIGHT` orientation)
   - one for vertical transformation (`BOTTOM` orientation)

flo-dup:
All calls to `Node::setRotationAngle` should then be replaced by `Node::setOrientation` (90 -> LEFT,  270 -> RIGHT, 180 -> BOTTOM, null -> TOP). 

So far, the nodes in a `BOTTOM` direction cell are not all rotated, except transformers. This should be fixed:
- in  `BodyPrimaryBlock::coordVerticalCase`, it should not always be set to null, but to TOP or BOTTOM according to block orientation.
- the code handling transformers rotation should be deleted
     - for transformers within a cell, they will get the orientation like the other nodes
     - for transformers outside cells (on a snakeline), the orientation should be set when calculating the snakelines, not afterwards in `handleMultiTermsNodeRotation`.

PR #377, opened 2022-03-23
Flip horizontally components in bottom direction cells

tadam50:

#191

feature

Components can be rotated when being in bottom direction cells but not flipped horizontally

Components can be easily flipped horizontally if in bottom direction cells

**Other information**:


single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractLayout.java
```diff
@@ -42,6 +44,12 @@ protected void manageSnakeLines(BaseGraph graph, LayoutParameters layoutParamete
                     ((BranchEdge) adjacentEdges.get(i)).setSnakeLine(pollingSplit.get(i));
                 }
             }
+
+            if (multiNode.getAdjacentEdges().size() == 2) {  // 2 windings transformer
+                handle2wtNodeRotation(multiNode);
+            } else {  // 3 windings transformer
+                handle3wtNodeRotation(multiNode);
+            }
```
flo-dup, line 52: I think you can put this two in the for loop above, to avoid the if again, and put the `List<List<Point>>` as argument instead of retrieving it

flo-dup (approved):

PR #452, opened 2022-11-25
[nad] Improve diagram metadata

flo-dup:

No

Code simplification

Metadata formats coordinates according to default locale `Locale`.
Metadata objects are the same for BusNode and Node, whereas they may contain different attributes.
When writing metadata objects, one needs to know whether they have already been prefixed or not.
metadata package has a dependence on other packages

Metadata coordinates are `String`, hence already formatted.
One NodeMetadata object, one BusNodeMetadata object
Metadata objects contain the svgId, not the diagramId, hence not wondering what's in prefix.
metadata package has no dependence on other packages

No


zamarrenolm (approved):
the changes left the code much cleaner and easier to read

PR #457, opened 2022-12-19
Edge parts can be stretchable or glued

zamarrenolm:
Signed-off-by: Luma <zamarrenolm@aia.es>

Feature

Some parts of the edge drawings are marked as `stretchable` or `glued`.
This information helps to update the diagram when the user interacts with it. 
In the basic use-case is a user wants to move some node locations and see the effects (immediately) in the diagram.
To provide this kind of interaction, we want to allow the user drag the nodes to their desired positions. The edges connected to the moved node must be updated (if possible, while the user is performing the drag operation).
To be able to properly update the different parts of the edge we label them using two classes:

- `stretchable`: this part of the edge drawing can be scaled to fill the new gap between end nodes.
- `glued`: this part of the edge is glued to one of the ends of the edge or to its center.

**Other information**:
Previously the symbol for the transformer was split in two drawings, each one contained in its half edge.
With this change the circle drawings for the transformer symbol are grouped together in an element that then is glued to center of the edge.

### DIFF
--- a/network-area-diagram/src/main/java/com/powsybl/nad/svg/StyleProvider.java
+++ b/network-area-diagram/src/main/java/com/powsybl/nad/svg/StyleProvider.java
@@ -39,6 +39,9 @@ public interface StyleProvider {
     String LABEL_BOX_CLASS = CLASSES_PREFIX + "label-box";
     String LEGEND_SQUARE_CLASS = CLASSES_PREFIX + "legend-square";
     String PST_ARROW_CLASS = CLASSES_PREFIX + "pst-arrow";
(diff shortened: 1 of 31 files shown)


flo-dup (approved):
I just added some minor modifications:
- drawEdgeInfo should always add the edge info style, therefore the styles passed as parameters are only additional styles
- it makes more sense to put the pst arrow in the same svg group as the windings
- there was one extra svg group in each winding to add the glued center class, but that class can be concatenated to the other classes of the winding
- the converter station has been forgotten

PR #459, opened 2022-12-28
add glued class to half edge drawings

zamarrenolm:

Feature

Currently the drawings for half edges are only labelled as _stretchable_.
To handle updates of these drawings during user interaction (moving nodes) we need to label each edge half as _glued_ to the corresponding edge end, in addition to _stretchable_.

Every half edge is labelled as _stretchable_ and _glued_ to its corresponding end.


flo-dup (approved):
This highlights the issue of the glued/stretchable binary approach: here the half edge has both the glued and stretchable class, so we cannot move it with a transform operation. I think we can't avoid this part of the half edge to be treated specifically in the js code. 

Anyway this is needed for moving the nodes in the js viewer with current approach, so approving this so that we can end and merge the viewer js part.

PR #460, opened 2022-12-29
draw external part of boundary nodes

zamarrenolm:

Feature

Only the _network_ part of boundary nodes has a drawing.

In addition to the drawing representing the inner (_network_) part of the boundary node, the external part is drawn using a dotted semi circumference. This is interesting because now the center of the boundary drawing coincides with the location of the boundary node computed by the layout.


flo-dup:
I think there's no need to change the design of this "node", as we agreed on the previous one. I understand that it's a trick to help rotating it but we'll give it a try with current design once everything else is fixed. Closing this.

PR #461, opened 2023-01-04
Display line names on diagram

### LINKED ISSUE #437: Add name on the lines
Feature

Lines are not identified

Lines identified by their name

Understanding better the network

So-Fras:

fixes #437


So-Fras:
comparisonDisplayLineNameOrNot

Line names are displayed along the lines (on the example picture, line names are "l1" and "l2").

network-area-diagram/src/main/java/com/powsybl/nad/svg/StyleProvider.java
```diff
@@ -27,6 +27,7 @@ public interface StyleProvider {
     String DANGLING_LINE_EDGE_CLASS = CLASSES_PREFIX + "dangling-line-edge";
     String TEXT_EDGES_CLASS = CLASSES_PREFIX + "text-edges";
     String EDGE_INFOS_CLASS = CLASSES_PREFIX + "edge-infos";
+    String EDGE_ID_CLASS = CLASSES_PREFIX + "edge-id";
```
flo-dup, line 30:
```suggestion
    String EDGE_LABEL_CLASS = CLASSES_PREFIX + "edge-label";
```

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -129,13 +129,57 @@ private void drawBranchEdges(Graph graph, XMLStreamWriter writer) throws XMLStre
...
+            if (!edge.isVisible(BranchEdge.Side.ONE)) {
+                edgeAngle = edge.getEdgeEndAngle(BranchEdge.Side.TWO);
+                textFlipped = Math.cos(edgeAngle) < 0;
+                style = textFlipped ? "text-anchor:end" : "text-anchor:start";
+            } else if (!edge.isVisible(BranchEdge.Side.TWO)) {
+                style = textFlipped ? "text-anchor:end" : "text-anchor:start";
+            }
+
+            writer.writeStartElement(GROUP_ELEMENT_NAME);
+            writer.writeAttribute(CLASS_ATTRIBUTE, StyleProvider.EDGE_ID_CLASS);
```
flo-dup, line 174: Add the `GLUED_CENTER_CLASS` at the same time, using `writeStyleClasses` method.

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -162,6 +206,7 @@ private void draw2Wt(XMLStreamWriter writer, BranchEdge edge) throws XMLStreamEx
         if (BranchEdge.PST_EDGE.equals(edge.getType())) {
             drawPstArrow(writer, edge);
         }
+
```
flo-dup, line 209: To avoid spoiling the history you'd better remove formatting changes (except if you changed code next to it).

flo-dup (changes requested):
Nice! just a few changes needed, in order not to break the interaction feature and for users to be able to customize what is displayed.

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -129,13 +129,57 @@ private void drawBranchEdges(Graph graph, XMLStreamWriter writer) throws XMLStre
...
             writer.writeEndElement();
         }
         writer.writeEndElement();
     }
 
+    private void drawLineLabel(XMLStreamWriter writer, BranchEdge edge) throws XMLStreamException {
+
+        String lineName = edge.getEquipmentId();
+
+        if (lineName != null) {
```
flo-dup, line 154: You could also test if `lineName.isEmpty()`.
And to cut complexity, you'd better return if null or empty.

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -129,13 +129,57 @@ private void drawBranchEdges(Graph graph, XMLStreamWriter writer) throws XMLStre
...
+
+            if (BranchEdge.PST_EDGE.equals(edge.getType())) {
+                drawPstArrow(writer, edge);
+            }
+            if (edge.getType().equals(BranchEdge.HVDC_LINE_EDGE)) {
+                drawConverterStation(writer, edge);
+            }
+            if (BranchEdge.LINE_EDGE.equals(edge.getType()) && svgParameters.isLineNameDisplayed()) {
+                drawLineLabel(writer, edge);
+            }
```
flo-dup, line 141: This should be (and is already partly, leading to duplicates) in drawEdgeCenter

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -129,13 +129,57 @@ private void drawBranchEdges(Graph graph, XMLStreamWriter writer) throws XMLStre
...
             drawHalfEdge(graph, writer, edge, BranchEdge.Side.ONE);
             drawHalfEdge(graph, writer, edge, BranchEdge.Side.TWO);
+
+            if (BranchEdge.PST_EDGE.equals(edge.getType())) {
+                drawPstArrow(writer, edge);
+            }
+            if (edge.getType().equals(BranchEdge.HVDC_LINE_EDGE)) {
+                drawConverterStation(writer, edge);
+            }
+            if (BranchEdge.LINE_EDGE.equals(edge.getType()) && svgParameters.isLineNameDisplayed()) {
```
flo-dup, line 139: Did you try to display the name also above two-winding transformers, phase-shift transformers and converter stations?

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
So-Fras: In order to take into account another of your points, I would like to call the `drawLineLabel(...)` inside the `drawEdgeCenter(...)` function, where there is already a `writer.writeAttribute(CLASS_ATTRIBUTE, StyleProvider.GLUED_CENTER_CLASS)`.
I was thinking of not adding the GLUED_CENTER_CLASS to the drawLineLabel(...) function. 
What do you think about that ?

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
So-Fras: Displaying names above phase-shift transformers, converter stations and two-winding transformers does not look good because of the drawing already present at the center of the edge. I was thinking of writing the names slightly above the edge to avoid this but when there are several lines carrying two-winding transformers for example running in parallel, the problem remains. It looks a bit messy.

So-Fras: As discussed this morning, the diagram remains readable with names on every type of edge. Since this piece of information seems meaningful for users, we finally choose to display names, regardless of the edge type. The HVDC converter station design is slightly modified to be more visible even with the edge name displayed.

flo-dup (changes requested):

flo-dup (approved):
Great! It's a lot clearer that way :+1:

PR #462, opened 2023-01-09
Use min value instead of average value to order bus cells

So-Fras:

Feature

The position of the cell on the bus bar is determined by the average position numbers of its components.

The position of the cell on the bus bar is determined by the minimum position number among its components.


flo-dup (commented):
Looking at previous lines below, it seems that they are unneeded, maybe we can take the opportunity to remove them
https://github.com/powsybl/powsybl-diagram/blob/2dfc72b545a83161cd65ddabac752a80da49f960/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/PositionFromExtension.java#L120-L124

flo-dup:
In fact the mentioned lines should be removed but I forgot one thing: you should set the direction of the bus cell inside the `graph.getBusCellStream().forEach()` part. Currently, in the mentioned lines, the last node in the stream decide the direction if there are several different directions, and this is bad. You should check that the directions are coherent, and log a warning (throw an exception is maybe too harsh) if they're not.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/PositionFromExtension.java
```diff
@@ -117,17 +117,26 @@ public LBSCluster organizeLegBusSets(VoltageLevelGraph graph, List<LegBusSet> le
...
-            cell.setDirection(node.getDirection());
-            node.getOrder().ifPresent(cell::setOrder);
-        });
+
         graph.getBusCellStream().forEach(bc -> {
+            List<Direction> listOfDirectionsInsideCell = bc.getNodes().stream().map(Node::getDirection).distinct().collect(Collectors.toList());
+            int numberOfDirectionsInsideCell = listOfDirectionsInsideCell.size();
+            if (numberOfDirectionsInsideCell == 1) {
+                Direction direction = listOfDirectionsInsideCell.get(0);
+                if (direction != Direction.UNDEFINED) {
```
flo-dup, line 126: you can remove that test: `BusCell` direction is `UNDEFINED` by default, so `bc.setDirection(UNDEFINED)` is not very harmful!

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/PositionFromExtension.java
```diff
@@ -117,17 +117,26 @@ public LBSCluster organizeLegBusSets(VoltageLevelGraph graph, List<LegBusSet> le
...
+                    bc.setDirection(direction);
+                }
+            } else if (numberOfDirectionsInsideCell > 1) {
+                bc.setDirection(listOfDirectionsInsideCell.get(numberOfDirectionsInsideCell - 1));
```
flo-dup, line 130: I'd say `get(0)` instead, or is there any reason to choose last one rather than first one?

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/PositionFromExtension.java
So-Fras: I chose the last one so that, if we ever fall into this case, we would not differ from what was done before.
But if you think it is more logical to get the first one, it is fine for me too !

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/PositionFromExtension.java
So-Fras: If I remove the condition, some tests fail. There might be an underlying bug in the setDirection function. I need to check that further!

flo-dup (approved):

Issue #474, opened 2023-01-12
Improve code coverage


So-Fras:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Some parts of the powsybl-diagram code have a very low test coverage.

* **What is the expected behavior?**
The first task is to assess which parts of the code are lacking tests.
Tests should be added accordingly.

* **What is the motivation / use case for changing the behavior?**
Improve reliability

PR #482, opened 2023-01-23
Improve code coverage

tadam50:

#474


So-Fras (changes requested):
Thank you for your work! I made some minor comments.
Regarding the NetworkAreaDiagramTest, I think we should maybe merge the test as it is and change it after the API modification (which is not completed yet).
Is there an easy way to assess the change on code coverage % before merging ?

So-Fras (approved):

flo-dup (approved):
Thank you for this meticulous work!

Issue #471, opened 2023-01-11
Make undefined value symbol customisable


So-Fras:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
When a value is undefined, a dash (i.e. a "long hyphen") is displayed in diagrams.

* **What is the expected behavior?**
We want to make the "undefined value" symbol customisable.

* **What is the motivation / use case for changing the behavior?**
Add more flexibility for users.

Issue #469, opened 2023-01-11
Make SLD and NAD APIs more consistent


So-Fras:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
SLD and NAD APIs are different.
For example, in order to draw a diagram with SLD, the following code is used:

```
SingleLineDiagram.draw(...);
```

When generating a diagram with NAD, one has to do the following:

```
NetworkAreaDiagram nad = new NetworkAreaDiagram(...);
nad.draw(...);
```

* **What is the expected behavior?**
Both APIs should work the same for the draw() function and be as consistent as possible for the rest of the code.

* **What is the motivation / use case for changing the behavior?**
This will help users to master both APIs more easily.

* **Please tell us about your environment:**
  - PowSyBl Version: 3.0.0

PR #483, opened 2023-01-26
Undefined value symbol as a parameter (instead of an em dash constant)

So-Fras:

#471

**Other information**:
In SLD, the valueFormatter (where the "undefined value symbol" is set as a parameter) is created in the LayoutParameters class.  In a future release, a new SvgParameter class will be created (see #469) for Single Line Diagram. The valueFormatter creation shall then be moved in the newly created class

### DIFF
--- a/network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
+++ b/network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
@@ -53,8 +53,6 @@ public class SvgParameters {
     private int powerValuePrecision = 0;
     private int angleValuePrecision = 1;
     private double pstArrowHeadSize = 8;
-
-    /** em dash unicode for undefined value */
     private String undefinedValueSymbol = "";
 
     public enum CssLocation {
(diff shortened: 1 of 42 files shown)


network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
```diff
@@ -54,6 +54,9 @@ public class SvgParameters {
     private int angleValuePrecision = 1;
     private double pstArrowHeadSize = 8;
 
+    /** em dash unicode for undefined value */
+    private String undefinedValueSymbol = "\u2014";
```
flo-dup, line 58: Could you replace this by `""`? it doesn't really look good a dash above an edge, and now we can take advantage of this new feature!

flo-dup (commented):
Great! juste one minor comment

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
So-Fras: Yes, my bad, I forgot we talked about that !

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
```diff
@@ -54,6 +54,9 @@ public class SvgParameters {
     private int angleValuePrecision = 1;
     private double pstArrowHeadSize = 8;
 
+    /** em dash unicode for undefined value */
```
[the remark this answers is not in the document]
So-Fras: yes indeed...

flo-dup (approved):

PR #486, opened 2023-01-27
get number of voltage levels in nad diagram

FranckLecuyer:

Feature

**Other information**:


flo-dup (commented):
Just wondering why you need this, the number of nodes in the metadata could be used instead, but you want this information before generating the SVG is that it?

FranckLecuyer:
> Just wondering why you need this, the number of nodes in the metadata could be used instead, but you want this information before generating the SVG is that it?

We need this in gridstudy to prevent increasing the nad depth when a fixed number of voltage levels has been reached.
The diagram metadata is not accessible from the NetworkAreaDiagram class, so i decided to use the voltageLevelFilter to get this information ...

flo-dup (approved):
I removed the method from `NetworkAreaDiagram` as this class is probably going to be a static final class in next release. You can still access to the number of voltage levels by 
- either creating the voltageLevelFilter on your side and call `getNbVoltageLevels()` on it
- or calling `getVoltageLevelCount()` on the network

Issue #470, opened 2023-01-11
Layout instability


So-Fras:
* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
When drawing 2 diagrams with the same network, the layout may be different.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
The bug happens when drawing 2 diagrams from the same network. The equipment lists are not necessarily read in the same order and thus the layouts may be different from one diagram to the other.

* **What is the expected behavior?**
Same layouts for both diagrams

* **What is the motivation / use case for changing the behavior?**
Not confusing the user

* **Please tell us about your environment:**
  - PowSyBl Version: 3.0.0

PR #490, opened 2023-01-29
Fix nad layout instability

So-Fras:

#470


network-area-diagram/src/main/java/com/powsybl/nad/build/iidm/VoltageLevelFilter.java
```diff
@@ -71,6 +73,14 @@ public static VoltageLevelFilter createVoltageLevelsFilter(Network network, List
         return createVoltageLevelsDepthFilter(network, voltageLevelIds, 0);
     }
 
+    public static Collection<VoltageLevel> getNextDepthVoltageLevels(Network network, List<VoltageLevel> voltageLevels) {
+        List<String> voltageLevelIds = voltageLevels.stream().map(VoltageLevel::getId).collect(Collectors.toList());
+        VoltageLevelFilter voltageLevelFilter = createVoltageLevelsDepthFilter(network, voltageLevelIds, 1);
+        Set<VoltageLevel> voltageLevelSet = new HashSet<>(voltageLevelFilter.getVoltageLevels());
+        voltageLevels.forEach(voltageLevelSet::remove);
```
flo-dup, line 80: I think we can do better to avoid creating those sets. Maybe by using `traverseVoltageLevels` at the cost of changing its signature. Or maybe by enriching VoltageLevelFilter API, like in #491. But let's do that later in a separate PR

flo-dup (approved):

PR #492, opened 2023-01-30
[SLD] Fix set direction in bus cells

### LINKED ISSUE #485: Investigation needed on apparently unnecessary if condition
bug / investigation needed

In SLD, in the PositionFromExtension.java > gatherLayoutExtensionInformation () function, a test on the direction value is added before setting the bus cell direction:

```
if (direction != Direction.UNDEFINED) {
    bc.setDirection(direction);
}
```

If this condition is removed, some unit tests fail (there are errors in the drawings of intern cells).
However, this condition should not be necessary as the UNDEFINED value is the default value for direction.

In order to reproduce the bug, remove the if condition

The expected behavior is that the "if condition" is not necessary.  Some investigations are needed here to understand if there is an underlying bug.

Cleaner code

  - PowSyBl Version: 3.0.0

flo-dup:

Yes, fixes #485

Bug fix / code clean

Warning on almost all extern cells, and twisted code when setting directions in PositionFromExtension

No warning if two directions in an extern cell if one of them is `UNDEFINED`, clarify code by refactoring and adding comment

No


So-Fras (approved):
Thank you for solving issue #485

PR #496, opened 2023-01-31
NAD : fix dangling line drawing disconnected status

zamarrenolm:

Bug fix

Drawings of disconnected dangling lines are painted with one half dashed and one half solid.
See the problem here:
disconnected_bad

A disconnected dangling line should be painted with both halves using the same style.
Solid when it is connected and dashed when it is disconnected.
See the fixed drawing for disconnected status:
disconnected_ok

This is the (correct) drawing when the dangling line is connected:
connected_ok

### DIFF
--- a/network-area-diagram/src/test/java/com/powsybl/nad/svg/DanglingLineTest.java
+++ b/network-area-diagram/src/test/java/com/powsybl/nad/svg/DanglingLineTest.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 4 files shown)


network-area-diagram/src/test/java/com/powsybl/nad/svg/DanglingLineTest.java
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright (c) 2022, RTE (http://www.rte-france.com)
```
flo-dup, line 2:
```suggestion
 * Copyright (c) 2023, RTE (http://www.rte-france.com)
```

flo-dup (approved):
Well there should be only one dashed line and not two, but that would need some deeper changes - we'll fix that later.

Issue #237, opened 2021-05-07
Animate feeder arrows


flo-dup:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
Static feeder arrows

* **What is the expected behavior?**
Animated feeder arrows, if corresponding layout parameter enabled (with speed depending on the power flow values?)

* **What is the motivation / use case for changing the behavior?**
Nice looking? SVG. Not completely convinced of result so far (see below), a bit irritating instead.

* **Other information**
Simply adding the following lines to the components.css does the job, but we could go further to have a speed depending on flow values. With a file like the base-voltages.yml to define the 3 (for instance) speeds with min and max values.
```
.sld-arrow-in {
  offset-rotate: 0deg;
  offset-path: path('M 0,-3 0,3 0,-3');
  animation: move 2s linear infinite;}
.sld-arrow-out {
  offset-rotate: 0deg;
  offset-path: path('M 0,3 0,-3 0,3');
  animation: move 2s linear infinite;}
@keyframes move {
  to { 
    offset-distance: 100%;
  }
}
```
The CSS properties needed are still marked as experimental, but work in most browsers, see [browser compatibility sheet](https://developer.mozilla.org/en-US/docs/Web/CSS/offset-path#browser_compatibility).
 
![animatedArrows](https://user-images.githubusercontent.com/66690739/117458621-c4da1780-af4a-11eb-99c1-fc91f8cdb7a4.gif)
(here in gif, as github doesn't support svg files - result's smoother in svg)

So-Fras:
- It would be better if the arrows did not go back and forth. The move should only be one-way (in the direction of the arrow);
- The user should have the possibility to switch this feature on and off.

PR #500, opened 2023-02-13
[SLD] 237 animate feeder arrows

tadam50:

#237


tadam50:
Breaking changes is raised here by adding new parameter into 
`public List<String> getFeederInfoStyles(FeederInfo info, boolean animated)`
This new parameter is used to give `LayoutParameter::isEnableFeederInfosAnimation()`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DiagramStyles.java
```diff
@@ -46,6 +46,14 @@ public final class DiagramStyles {
     public static final String ANGLE = STYLE_PREFIX + "angle";
     public static final String FEEDER_INFO = STYLE_PREFIX + "feeder-info";
 
+    public static final String ARROW_ANIMATION = STYLE_PREFIX + "arrow-animation";
+
+    public static final String ARROW_ANIMATION_SPEED_1 = STYLE_PREFIX + "arrow-animation-speed-1";
```
So-Fras, line 51: We have to decide how to discretize the power values and maybe include  "no value" or "zero power" cases.

So-Fras (changes requested):
Thank you for your work so far. We must clear up some specifications for you to be able to go on with the development.
Moreover, we did not realize at first that this feature could bring a breaking change.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/util/AnimatedFeederInfoStyleProvider.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class AnimatedFeederInfoStyleProvider extends TopologicalStyleProvider {
+
+    private static final String ARROW_ANIMATION = STYLE_PREFIX + "arrow-animation";
+
+    private static final String ARROW_SPEED = "speed";
+
+    private static final String ARROW_ANIMATION_LOW_SPEED = ARROW_ANIMATION + "-low-" + ARROW_SPEED;
+
+    private static final String ARROW_ANIMATION_NOMINAL_SPEED = ARROW_ANIMATION + "-nominal-" + ARROW_SPEED;
```
So-Fras, line 28: I would not call it "nominal" but "average" maybe.

single-line-diagram/single-line-diagram-core/src/main/resources/animations.css
```diff
@@ -0,0 +1,32 @@
...
+  animation: move var(--sld-arrow-animation-duration, 0s) linear infinite;
+}
+@keyframes move {
+  to {
+    offset-distance: 100%;
+  }
+}
+/* Set arrows animation speed (1: the slowest -> 3: the fastest) */
+.sld-arrow-animation-low-speed { --sld-arrow-animation-duration: 3s }
+.sld-arrow-animation-nominal-speed { --sld-arrow-animation-duration: 2s }
```
So-Fras, line 31: If you change the name from "nominal" to "average", it will have an impact here too!

So-Fras (changes requested):
I made very few comments. Thank you for your nice work!

So-Fras (approved):
Everything is OK for me!

single-line-diagram/single-line-diagram-core/src/test/resources/TestAnimatedFeederInfos.svg
flo-dup, line 1: This file is not working on my firefox 96.0, the bottom black arrow is pointing to the top instead of to the bottom. It seems that the rotate is overriden. But in [the specifications](https://drafts.fxtf.org/motion-1/#offset-rotate-property) it is written that:
> Note: The rotation described here does not override or replace any rotation defined by the [transform](https://drafts.csswg.org/css-transforms-1/#propdef-transform) property.

So, if it still occurs on a more recent version of firefox, we should report it to mozilla bug report.

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/test/resources/TestAnimatedFeederInfos.svg
tadam50: On Firefox ESR 102.9.0 (March 14, 2023) same issue

flo-dup (commented):
Just 2 typos and we're good!

flo-dup (approved):

PR #501, opened 2023-02-20
Remove fictitious switch nodes

So-Fras:

Feature

Fictitious nodes are always displayed. There is no choice left to the user.

The user is able to set a boolean to display switch nodes or not in single-line diagrams.  By default, fictitious switch nodes are displayed.

No

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/GraphRefiner.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/GraphRefiner.java
@@ -23,10 +23,12 @@ import java.util.stream.Collectors;
 public class GraphRefiner {
     private final boolean removeUnnecessaryFictitiousNodes;
     private final boolean substituteSingularFictitiousByFeederNode;
+    private final boolean removeFictitiousSwitchNode;
 
-    public GraphRefiner(boolean removeUnnecessaryFictitiousNodes, boolean substituteSingularFictitiousByFeederNode) {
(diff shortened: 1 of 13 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/GraphRefiner.java
```diff
@@ -23,10 +23,12 @@
 public class GraphRefiner {
     private final boolean removeUnnecessaryFictitiousNodes;
     private final boolean substituteSingularFictitiousByFeederNode;
+    private final boolean removeFictitiousSwitchNode;
```
flo-dup, line 26: In a way, it makes sense to put that parameter here, but how is it possible for the user to change this easily? It would be easier to put it in LayoutParameters. Does it make sense? I'd say yes

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/GraphRefiner.java
So-Fras: Isn't it ok to do:

```
 factory = new PositionVoltageLevelLayoutFactory();
 factory.setRemoveFictitiousSwitchNodes(true);
```

And then use this factory to draw the diagram?

So-Fras: I finally put it in LayoutParameters, it is indeed more user-friendly that way. It would be good too to think about moving other parameters from the PositionVoltageLevelLayoutFactory to the LayoutParameters. 
I created an issue for that #503.

flo-dup (approved):

PR #502, opened 2023-02-23
Simplify LegBusSet

BenoitJeanson:
no

Just simplify `LegBusSet `class:
- the `extendedNodeSet `is used only in `Substation`, so it is clearer to have it in the context of `Substation`
- `handleShunt `method was useless as `externCells.forEach(cell -> pushLBS(legBusSets, new LegBusSet(busToNb, cell)));` ends up with the same result


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LegBusSet.java
```diff
@@ -99,82 +86,36 @@ Set<InternCellSide> getInternCellSides() {
...
         allBusNodes.removeAll(legBusSets.stream().
                 flatMap(legBusSet -> legBusSet.getBusNodeSet().stream()).collect(Collectors.toList()));
         allBusNodes.stream()
                 .sorted(Comparator.comparing(Node::getId))              //avoid randomness
-                .forEach(busNode -> legBusSets.add(new LegBusSet(nodeToNb, busNode)));
+                .forEach(busNode -> legBusSets.add(new LegBusSet(busToNb, busNode)));
         return legBusSets;
     }
 
-    private static void manageShunts(VoltageLevelGraph graph, List<ExternCell> externCells, List<LegBusSet> legBusSets, Map<BusNode, Integer> nodeToNb) {
```
flo-dup, line 148: just tested on another branch (the one dealing with star shunts) with a more complex case, and in fact the result is not the same if removing this or not. It doesn't mean that this method is useful (the result with this manageShunts method is uglier in fact), but it seems that it wasn't doing the same as `externCells.forEach(cell -> pushLBS(legBusSets, new LegBusSet(busToNb, cell)));`. Is this normal?

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LegBusSet.java
So-Fras: It is unclear to me as well.

BenoitJeanson: I really think it should not change the expected behavior. It's ugly. Let's remove it.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/positionfromextension/HBLaneManagerByExtension.java
```diff
@@ -33,14 +29,13 @@ public void mergeHorizontalBusLanes(LBSCluster leftCluster, LBSCluster rightClus
...
                     .filter(hbl2 -> hbl2.getSideNode(Side.LEFT).getBusbarIndex() == rightNodeOfLeftHbl.getBusbarIndex())
                     .findFirst();
-            if (rightHBL.isPresent()
-                    && (rightHBL.get().getSideNode(Side.LEFT) == rightNodeOfLeftHbl
-                    || rightNodeOfLeftHbl.getSectionIndex() < rightHBL.get().getSideNode(Side.LEFT).getSectionIndex())) {
-                hbl.merge(rightHBL.get());
-                rightCluster.removeHorizontalBusLane(rightHBL.get());
-            } else {
-                // TODO : not true !!!
-                LOGGER.warn("incoherent structural horizontal positions in mergeHorizontalBusLanes");
```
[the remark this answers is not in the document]
BenoitJeanson: It should never happen (or it means the extension is not consistent), and if so mergeLanesWithNoLink should resolve it.

BenoitJeanson:
Will be held in another PR.

PR #515, opened 2023-04-17
Remove test jars usage

tadam50:

No

Technical debt

We use test jars in many places to share classes between different artifact unit tests.

We don't use test jars anymore but normal jar except for TCK. This will allow easier migration to Java module system.

**Other information**:


So-Fras (changes requested):
Nice and thorough work! I just have a minor remark.

So-Fras (approved):

PR #518, opened 2023-05-04
Using horizontal & vertical layouts for ZoneGraph

tadam50:

No

Feature

No layouts for Zone graph are available 

Provide horizontal and vertical layouts for Zone graph

**Other information**:


flieks:
This is great, thanks!

Just have 1 question. I see there is a horizontal example and vertical example. But can vertical and horizontal layouts be combined in 1 svg ?

For example we have 12 substations, then ideally we want to have 3 rows with each 4 substations...

So-Fras (changes requested):
Thank you for your nice work!
I didn't have time to finish my review, to be continued :)

So-Fras (changes requested):
Thank you for the huge amount of work you have put in this PR.
I made a minor comment, I would like to discuss it further with you because I do not get why there is a difference between substation and zone regarding layout naming (and furthermore I may be to blame here!).

So-Fras (approved):

So-Fras:
Some configurations (VH and HV) do not offer a clear and completly satisfying display. But this PR is a first attempt to display multi-substation graphs. It will be followed by another PR to represent multi-substation graphs according to a matrix design.

PR #521, opened 2023-05-25
Bus connection status representation bug

So-Fras:

No

Bug fix

In bus breaker view, there is no difference on the diagram if a node is connected to a bus bar or not. The bus connection is represented as a full circle in both cases.

**What is the new behavior?**
If the node is not connected to the bus bar, the bus connection is represented as an empty circle.


So-Fras:
The PR has a lot of conflicts to solve and can't be used as it is. however, it could be used for inspiration to improve the quick fix made in PR #530.

Issue #510, opened 2023-03-29
Clean cgmes-only parameters from LayoutParameters


So-Fras:
* **Do you want to request a *feature* or report a *bug*?**
Feature

* **What is the current behavior?**
In single-line diagram, some cgmes-only parameters (`cgmesUseName`, `cgmesScaleFactor`, `cgmesDiagramName`) are present in `LayoutParameters `but `LayoutParameters ` is not the right class to contain them. 
The rationale behind this was that the smart selector did not allow to pass any parameters, so the `LayoutParameters `class was used for cgmes-only parameters.

* **What is the expected behavior?**
Some refactoring is necessary to clean this. Cgmes-only parameters should not be stored in the `LayoutParameters `class.

samirromdhani:
The given cgmes-only parameters are already moved in `AbstractCgmesLayout`

-  `scaleFactor` done in #776 
- `cgmesScaleFactor`,`cgmesDiagramName`  done in #759

PR #522, opened 2023-06-05
Modify diagram apis

So-Fras:

#469
There is another PR on the same topic (https://github.com/powsybl/powsybl-diagram/pull/475). This old PR has diverged too much from main and it was easier to start afresh than to rebase or merge.
Developments on the old PR were added to this new PR. 
New developments have been added afterwards on the new branch: 
- Config and ConfigBuilder have been renamed in Param and ParamBuilder
- Param and ParamBuilder have become network agnostic, both for SLD and NAD (whereas it was only the case for NAD on the old PR)

Modification and standardization of SLD and NAD APIs

- [X] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration guide has been updated in the github wiki *(What changes might users need to make in their application due to this PR?)*


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagram.java
```diff
@@ -174,138 +128,71 @@ public static void draw(Graph graph, Path svgFile, LayoutParameters layoutParame
...
+        Objects.requireNonNull(substationLayoutFactory);
+        Objects.requireNonNull(voltageLevelLayoutFactory);
 
         SubstationGraph substationGraph = new NetworkGraphBuilder(network).buildSubstationGraph(substationId);
-        sLayoutFactory.create(substationGraph, vLayoutFactory).run(layoutParameters);
-        draw(substationGraph, writerForSvg, metadataWriter, layoutParameters, componentLibrary, initProvider, styleProvider, prefixId);
-    }
-
-    public static void draw(Graph graph, Writer writerForSvg, Writer metadataWriter, LayoutParameters layoutParameters, ComponentLibrary componentLibrary,
-                            DiagramLabelProvider initProvider, StyleProvider styleProvider, String prefixId) {
```
flo-dup, line 284: You removed this method, but it was used in unit tests. Removing it forced you to instantiate a DefaultSvgWriter in all the unit tests. We don't want this, as this leads to always the same code in all the tests.

With the new design the equivalent would be a `draw(graph, writerForSvg, metadataWriter, param)` method.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagram.java
```diff
@@ -174,138 +128,71 @@ public static void draw(Graph graph, Path svgFile, LayoutParameters layoutParame
...
-        draw(graph, writerForSvg, metadataWriter, new DefaultSVGWriter(componentLibrary, layoutParameters),
-                initProvider, styleProvider, prefixId);
+        substationLayoutFactory.create(substationGraph, voltageLevelLayoutFactory).run(layoutParameters);
+        DefaultSVGWriter svgWriter = new DefaultSVGWriter(param.getComponentLibrary(), param.getLayoutParameters(), param.getSvgParameters());
+        draw(substationGraph, writerForSvg, metadataWriter, svgWriter, param.createLabelProvider(network), param.createStyleProvider(network), param.getSvgParameters().getPrefixId());
     }
 
-    public static void draw(Graph graph, Writer writerForSvg, Writer metadataWriter, DefaultSVGWriter svgWriter,
-                            DiagramLabelProvider initProvider, StyleProvider styleProvider, String prefixId) {
+    public static void draw(Graph graph, Writer writerForSvg, Writer metadataWriter, SVGWriter svgWriter, LabelProvider labelProvider, StyleProvider styleProvider, String prefixId) {
```
flo-dup, line 183: I think we should hide the svgWriter and therefore remove this method. Besides, exposing the providers is against the new design.

This a utility class, hence we want to keep methods simple (simpler than they were!), an advanced user who wants to use a custom SvgWriter can still do it, without using this class.

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase15GraphWithVoltageIndicator.java
```diff
@@ -53,54 +55,66 @@ public boolean isPowered() {
...
-                        result = new BusVoltageInfo(true, "Top", null, Side.RIGHT);
-                        break;
-                    case "bbs22":
-                    case "bbs23":
-                        result = new BusVoltageInfo(false, null, "Bottom", Side.LEFT);
-                        break;
-                }
-                return Optional.ofNullable(result);
+            public LabelProvider create(Network network, ComponentLibrary componentLibrary, LayoutParameters layoutParameters, SvgParameters svgParameters) {
+                return new DefaultLabelProvider(network, componentLibrary, layoutParameters, svgParameters) {
```
flo-dup, line 70: you're creating here a new `DefaultLabelProviderFactory`, overriding the method create. This is an anonymous inner class which extends `DefaultLabelProviderFactory`. But there's nothing left of `DefaultLabelProviderFactory` as you override its only method. The benefit of a functional interface is that you can define it with a lambda, which leads to a more readable code, see below.

Could you simplify this? and also all the other anonymous inner classes you defined in unit tests.

```java
        withFullBusInfoProviderFactory = (network, componentLibrary, layoutParameters, svgParameters) ->
                new DefaultLabelProvider(network, componentLibrary, layoutParameters, svgParameters) {
                    @Override
                    public Optional<BusInfo> getBusInfo(BusNode node) {
                        (...)
                    }
                };
```

single-line-diagram/single-line-diagram-core/src/test/resources/TestSldClassSubstationMetadata.json
```diff
@@ -536,58 +536,36 @@
...
     "busbarsAlignment" : "FIRST",
     "componentsOnBusbars" : [ "DISCONNECTOR" ],
-    "languageTag" : "en",
-    "voltageValuePrecision" : 1,
-    "powerValuePrecision" : 0,
-    "angleValuePrecision" : 1,
-    "currentValuePrecision" : 0,
-    "displayCurrentFeederInfo" : false,
-    "undefinedValueSymbol" : "—",
-    "removeFictitiousSwitchNodes" : false
```
flo-dup, line 591: You've split the parameters in two parts, but you didn't add the `SvgParameters` in `GraphMetadata`. Are you sure they're of no use? I think we'd better add them in the metadata, so we keep the same information as before.

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/NetworkAreaDiagram.java
```diff
@@ -35,116 +31,91 @@
...
+        SvgParameters svgParameters = param.getSvgParameters();
+        StyleProvider styleProvider = param.createStyleProvider(network);
+        LayoutFactory layoutFactory = param.getLayoutFactory();
+        IdProvider idProvider = param.createIdProvider();
+        LabelProvider labelProvider = param.createLabelProvider(network);
         Objects.requireNonNull(layoutParameters);
         Objects.requireNonNull(svgParameters);
         Objects.requireNonNull(styleProvider);
         Objects.requireNonNull(layoutFactory);
         Objects.requireNonNull(idProvider);
```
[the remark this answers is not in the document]
So-Fras: In the Param constructor, those checks are on :
- layoutParameters (same as before)
- svgParameters (same as before)
- layoutFactory (same as before)
- styleProviderFactory (instead of styleProvider)
- idProviderFactory (instead of idProvider)

network-area-diagram/src/main/java/com/powsybl/nad/Param.java
```diff
@@ -0,0 +1,63 @@
...
+import com.powsybl.nad.layout.LayoutParameters;
+import com.powsybl.nad.svg.LabelProvider;
+import com.powsybl.nad.svg.StyleProvider;
+import com.powsybl.nad.svg.SvgParameters;
+
+/**
+ *
+ * @author Sophie Frasnedo <sophie.frasnedo at rte-france.com>
+ */
+public class Param {
```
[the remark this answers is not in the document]
So-Fras: While renaming only the Param class in NadParameters should be enough not to get confused, it seems more homogeneous to rename the single-line-diagram-core Param in SldParameters too. What do you think?

network-area-diagram/src/main/java/com/powsybl/nad/Param.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public SvgParameters getSvgParameters() {
+        return svgParameters;
+    }
+
+    public LayoutParameters getLayoutParameters() {
+        return layoutParameters;
+    }
+
+    public StyleProvider createStyleProvider(Network network) {
```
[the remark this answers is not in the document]
flo-dup: Following the discussion about labelProviderFactory, which makes the getter a bit difficult to handle, I think that, to have this logic applied to the labelProvider, we should have a simplified interface, for sld for instance:
```java
    @FunctionalInterface
    public interface LabelProviderFactory {
        LabelProvider create(Network network);
    }
```

And then the default value would be defined in the builder as:
```java
LabelProviderFactory labelProviderFactory = n -> new DefaultLabelProvider(n, componentLibrary, layoutParameters, svgParameters);
```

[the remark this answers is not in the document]
flo-dup: Ok, what I suggested isn't really a good idea, because when you define the label provider with the builder, it becomes something very clumsy:
```java
param = new ParamBuilder()
    .withLabelProviderFactory(n -> 
        new DefaultLabelProvider(n, componentLibrary, layoutParameters, svgParameters))
    .build();
```
Which forces you to define the componentLibrary, the layoutParameters and the svgParameters even if you want the default one.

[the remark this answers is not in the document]
So-Fras: Agreed. To me, there is no obvious solution to get rid of this createLabelProvider in the SldParameters and NadParameters classes.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagram.java
```diff
@@ -97,74 +79,46 @@ public static void drawVoltageLevel(Network network, String id, String svgFile)
...
-        Objects.requireNonNull(vLayoutFactory);
+        Objects.requireNonNull(substationLayoutFactory);
+        Objects.requireNonNull(voltageLevelLayoutFactory);
 
         SubstationGraph substationGraph = new NetworkGraphBuilder(network).buildSubstationGraph(substationId);
-        sLayoutFactory.create(substationGraph, vLayoutFactory).run(layoutParameters);
-        draw(substationGraph, svgFile, layoutParameters, componentLibrary, initProvider, styleProvider, prefixId);
+        substationLayoutFactory.create(substationGraph, voltageLevelLayoutFactory).run(layoutParameters);
+        DefaultSVGWriter svgWriter = new DefaultSVGWriter(param.getComponentLibrary(), param.getLayoutParameters(), param.getSvgParameters());
+        draw(substationGraph, svgFile, svgWriter, param.createLabelProvider(network), param.createStyleProvider(network), param.getSvgParameters().getPrefixId());
```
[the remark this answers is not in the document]
So-Fras: I removed this prefeixId parameter from several subfunctions in the DefaultSvgWriter as well, since it is contained in the metadata (already a parameter of those subfunctions) and that calling it from the metadata is quite simple (simple getters).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagramTool.java
```diff
@@ -86,63 +80,39 @@ public String getUsageFooter() {
     }
 
     private Path getSvgFile(Path outputDir, String id) {
-        try {
-            return outputDir.resolve(URLEncoder.encode(id, StandardCharsets.UTF_8.name()) + ".svg");
-        } catch (UnsupportedEncodingException e) {
-            throw new PowsyblException(e);
-        }
```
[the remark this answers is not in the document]
So-Fras: It was an IntelliJ suggestion to remove the try catch and replace the `StandardCharsets.UTF_8.name()` by `StandardCharsets.UTF_8`. I removed this change, it should be done in a separate PR.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagram.java
So-Fras: I did not wish to expose the providers but the LabelProvider and the StyleProvider cannot be retrieved using an SldParameters object only. I need a Network object too. That is why I had to "build" those providers earlier, when I could still have access to the Network object.

network-area-diagram/src/main/java/com/powsybl/nad/NadParametersBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+        this.layoutFactory = Objects.requireNonNull(layoutFactory);
+        return this;
+    }
+
+    public NadParametersBuilder withIdProviderFactory(IdProviderFactory idProviderFactory) {
+        this.idProviderFactory = Objects.requireNonNull(idProviderFactory);
+        return this;
+    }
+
+    public NadParameters build() {
```
geofjamg, line 60: An empty (no check needed) in the build shows that `NadParametersBuilder` class is useless. You should only rely a simple NadParameters bean with fields already set (like in the builder) and some additional setter to replace implementations. 
```NadParameters``` does not have to be immutable.

File: network-area-diagram/src/main/java/com/powsybl/nad/NadParameters.java [hunk 11]
```diff
@@ -0,0 +1,74 @@
...
+    public LayoutParameters getLayoutParameters() {
+        return layoutParameters;
+    }
+
+    public StyleProviderFactory getStyleProviderFactory() {
+        return styleProviderFactory;
+    }
+
+    public LabelProvider createLabelProvider(Network network) {
+        return labelProviderFactory.create(network, svgParameters);
```
geofjamg, line 55: Should should avoid this kind of method that does nothing really helpful
```java
parameters.getLabelProvider().create()
```
is enough.

network-area-diagram/src/main/java/com/powsybl/nad/NetworkAreaDiagram.java
```diff
@@ -35,116 +31,86 @@
...
-    public NetworkAreaDiagram(Network network, Predicate<VoltageLevel> voltageLevelFilter) {
-        this.network = Objects.requireNonNull(network);
-        this.voltageLevelFilter = Objects.requireNonNull(voltageLevelFilter);
+    private NetworkAreaDiagram() {
     }
 
-    public Network getNetwork() {
-        return network;
-    }
+    /* ------------------------------------------------------------------------------------------------ */
```
geofjamg, line 39: This is not standard Javadoc

single-line-diagram/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/main/java/com/powsybl/sld/cgmes/layout/LayoutToCgmesDlExporterTool.java
```diff
@@ -118,7 +118,7 @@ public void run(CommandLine line, ToolRunningContext context) {
         Network network = Network.read(inputFile);
 
         context.getOutputStream().println("Generating layout for the network ...");
-        LayoutToCgmesExtensionsConverter lTranslator = new LayoutToCgmesExtensionsConverter(sFactory, vFactory, new LayoutParameters().setUseName(true));
+        LayoutToCgmesExtensionsConverter lTranslator = new LayoutToCgmesExtensionsConverter(sFactory, vFactory, new LayoutParameters().setCgmesUseNames(true));
```
geofjamg, line 121: Why cgmesUseNames ? ID vs name is an IIDM concept, not only related to CGMES.

single-line-diagram/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/test/java/com/powsybl/sld/cgmes/layout/AbstractCgmesVoltageLevelLayoutTest.java
```diff
@@ -27,8 +27,8 @@ abstract class AbstractCgmesVoltageLevelLayoutTest {
     protected void test(VoltageLevel vl) {
         VoltageLevelGraph graph = new NetworkGraphBuilder(vl.getNetwork()).buildVoltageLevelGraph(vl.getId());
         LayoutParameters layoutParameters = new LayoutParameters();
-        layoutParameters.setScaleFactor(2);
-        layoutParameters.setDiagramName(DIAGRAM_NAME);
+        layoutParameters.setCgmesScaleFactor(2);
```
geofjamg, line 30: I'm not sure here too that we should call it CGMES. As it is true that today the extension (so no automatic) layout mode onyly comes from CGMES it is the plan to move all theses extension to core and not anymore related to CGMES.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SldParametersBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+import com.powsybl.sld.svg.styles.StyleProviderFactory;
+
+import java.util.Objects;
+
+/**
+ *
+ * @author Sophie Frasnedo <sophie.frasnedo at rte-france.com>
+ */
+
+public class SldParametersBuilder {
```
geofjamg, line 26: Same remark that for NadParameterBuilder. This is over-engineered. A simple SldParameters mutable bean with pre-initialized values would be a much simpler and user friendly API.
```java
new SldParameters()
    .setSvgParameters(new SvgParameters(...))
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
```diff
@@ -533,4 +536,10 @@ public LayoutParameters getLayoutParameters() {
         return layoutParameters;
 
     }
+
+    @JsonProperty("svgParams")
```
geofjamg, line 540: Why do you need this in addition to the constructor one?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/SvgParameters.java
```diff
@@ -0,0 +1,312 @@
...
+
+    public SvgParameters() {
+    }
+
+    public enum CssLocation {
+        INSERTED_IN_SVG, EXTERNAL_IMPORTED, EXTERNAL_NO_IMPORT
+    }
+
+    public SvgParameters(SvgParameters other) {
+        this.prefixId = other.prefixId;
```
geofjamg, line 53: A general comment: it is really important in an API to check for null parameters.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/SvgParameters.java
```diff
@@ -0,0 +1,312 @@
...
+        this.svgWidthAndHeightAdded = svgWidthAndHeightAdded;
+        return this;
+    }
+
+    public CssLocation getCssLocation() {
+        return cssLocation;
+    }
+
+    public SvgParameters setCssLocation(CssLocation cssLocation) {
+        this.cssLocation = cssLocation;
```
geofjamg, line 255: Also here for instance: null is an acceptable value?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase11SubstationGraph.java
```diff
@@ -138,11 +139,9 @@ void testHorizontalDefaultStyle() {
         // (with horizontal substation layout)
         SubstationGraph substationGraph = graphBuilder.buildSubstationGraph(substation.getId());
 
-        assertTrue(compareMetadata(substationGraph, "/substDiag_metadata.json",
-                new HorizontalSubstationLayoutFactory(),
-                new PositionVoltageLevelLayoutFactory(),
-                new DefaultDiagramLabelProvider(network, componentLibrary, layoutParameters),
-                new BasicStyleProvider()));
+        DefaultSVGWriter defaultSVGWriter = new DefaultSVGWriter(componentLibrary, layoutParameters, svgParameters);
```
geofjamg, line 142: This line of code is added everywhere. Wasn't it possible to create it in  compareMetadata for instance to keep code simple?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestSingleLineDiagramClass.java
```diff
@@ -80,7 +81,7 @@ void testSvgVl1() throws IOException {
         assertEquals(expected, toDefaultSVG(network, vl.getId(), "/TestSldClassVl.svg", "/TestSldClassVlMetadata.json"));
 
         Writer writerForSvg = new StringWriter();
-        SingleLineDiagram.drawVoltageLevel(network, vl.getId(), writerForSvg, new NullWriter());
+        SingleLineDiagram.drawVoltageLevel(network, vl.getId(), writerForSvg, new NullWriter(), SldParameters.defaultParameters());
```
geofjamg, line 84: new SldParameters() would be enough without this boilerplate builders

geofjamg (changes requested):

single-line-diagram/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/test/java/com/powsybl/sld/cgmes/layout/AbstractCgmesVoltageLevelLayoutTest.java
So-Fras: It is a workaround I found in order not to touch this part of the code in this PR. I created an [issue](https://github.com/powsybl/powsybl-diagram/issues/510) to track that.

single-line-diagram/single-line-diagram-cgmes/single-line-diagram-cgmes-layout/src/main/java/com/powsybl/sld/cgmes/layout/LayoutToCgmesDlExporterTool.java
So-Fras: It is a workaround I found in order not to touch this part of the code in this PR. I created an https://github.com/powsybl/powsybl-diagram/issues/510 to track that.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/SvgParameters.java
So-Fras: I added a non null check

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SldParametersBuilder.java
So-Fras: Done :)

network-area-diagram/src/main/java/com/powsybl/nad/NadParametersBuilder.java
So-Fras: Done :)

network-area-diagram/src/main/java/com/powsybl/nad/NadParameters.java
So-Fras: This method was created so that the svgParameters call is masked and dealt with inside the NadParameters class. There is not any possibility for a user to feed the create function with a SvgParameters that is not the 
It "replaces" the getter. There is no getter for a LabelProviderFactory or a LabelProvider in the NadParameters class (it may be wrong :)). 
What do you think?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase11SubstationGraph.java
So-Fras: I removed those lines. The `DefaultSvgWriter` is now created in a draw function, in the `SingleLineDiagram` class.

geofjamg (approved):

PR #523, opened 2023-06-06
Bump to 5.3.0

So-Fras:

Bump to new powsybl-core version.
This includes the management of TieLines for NAD and SLD

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
@@ -374,11 +374,14 @@ public class NetworkGraphBuilder implements GraphBuilder {
             if (!dl.isPaired()) {
                 addTerminalNode(NodeFactory.createDanglingLine(graph, dl.getId(), dl.getNameOrId()), dl.getTerminal());
             } else {
-                TieLine tieLine = dl.getTieLine().get();
-                Branch.Side side = tieLine.getDanglingLine1().getId().equals(dl.getId()) ? Branch.Side.ONE : Branch.Side.TWO;
-                Terminal terminal = dl.getTerminal();
-                addTerminalNode(createFeederTieLineNode(graph, tieLine, side), terminal);
+                dl.getTieLine().ifPresent(tieLine -> visitTieLine(tieLine, dl, graph));
             }
+        }
+
+        private void visitTieLine(TieLine tieLine, DanglingLine dl, Graph graph) {
+            Branch.Side side = tieLine.getDanglingLine1().getId().equals(dl.getId()) ? Branch.Side.ONE : Branch.Side.TWO;
+            Terminal terminal = dl.getTerminal();
+            addTerminalNode(createFeederTieLineNode((VoltageLevelGraph) graph, tieLine, side), terminal);
         }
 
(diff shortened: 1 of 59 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
```diff
@@ -346,7 +371,17 @@ public void visitShuntCompensator(ShuntCompensator sc) {
...
         public void visitDanglingLine(DanglingLine dl) {
-            addTerminalNode(NodeFactory.createDanglingLine(graph, dl.getId(), dl.getNameOrId()), dl.getTerminal());
+            if (!dl.isPaired()) {
+                addTerminalNode(NodeFactory.createDanglingLine(graph, dl.getId(), dl.getNameOrId()), dl.getTerminal());
+            } else {
+                dl.getTieLine().ifPresent(tieLine -> visitTieLine(tieLine, dl, graph));
+            }
+        }
+
+        private void visitTieLine(TieLine tieLine, DanglingLine dl, Graph graph) {
```
annetill, line 381: To be checked with @flo-dup and @miovd but it seems that we have nothing that prevents two dangling lines, connected to the same voltage level and paired together. It is a "bug" of usage and I think that in the case, the user wants to see the wierd tie line. I wonder if we are supporting that here.

annetill: It seems that you are supporting that looking at the last picture of your PR!

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
```diff
@@ -632,6 +667,27 @@ private void addLineEdges(Graph graph, List<Line> lines) {
         });
     }
 
+    private void addTieLineEdges(Graph graph, List<TieLine> tieLines) {
```
flo-dup, line 670: Duplicate code here which can be easily extended, as `TieLine` and `Line` extends `Branch` which is enough to what's done inside. Hence you could add a `addBranchEdges(Graph graph, List<? extends Branch<?>> branches)` method

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultDiagramLabelProvider.java
```diff
@@ -189,6 +189,22 @@ private void addBranchStatusDecorator(List<NodeDecorator> nodeDecorators, Node n
         }
     }
 
+    private void addBranchStatusDecorator(List<NodeDecorator> nodeDecorators, Node node, Direction direction, Branch<?> branch) {
```
flo-dup, line 192: This can be removed, `BranchStatus` is a `Connectable` extension, despite its name. So, `addBranchStatusDecorator` has to be called on feederTypes branches/2wt only if `network.getBranch(feederNode.getEquipmentId())` returns a connectable.

But, if it is a TieLine, we should call it on the corresponding dangling line. Let's keep that for later?

flo-dup (changes requested):

flo-dup (approved):

PR #528, opened 2023-07-03
Update to Java 17

So-Fras:

No

Update to Java 17
A test had to be modified in order to pass when run with Java 17 (French thousand separator was changed from Non-breakable space - NBSP - to Narrow non-breakable space - NNBSP - in Java 13 due to [CLDR 35.1 support](https://bugs.openjdk.org/browse/JDK-8221432)).

Yes


So-Fras:
Now I think about it, is there a reason to have a conditional test for the Java version? Once we switch to Java 17, will users be able to run tests with older Java versions anyway?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestFeederInfos.java
```diff
@@ -136,7 +136,7 @@ void testFrenchFormatting() {
         voltageLevelGraphLayout(g);
 
         // write SVG and compare to reference
-        assertEquals(toString("/TestFormattingFeederInfos.svg"), toSVG(g, "/TestFormattingFeederInfos.svg"));
+        assertEquals(toString("/TestFormattingFeederInfosNNBSP.svg"), toSVG(g, "/TestFormattingFeederInfos.svg"));
```
olperr1, line 139: Since we finally have only one version of the reference file, we can keep its original name (`TestFormattingFeederInfos.svg`).

olperr1 (approved):

PR #540, opened 2023-07-27
536 viewbox incorrectly specified in some cases depending on the direction of a line between two voltage levels

### LINKED ISSUE #536: Viewbox incorrectly specified in some cases depending on the direction of a line between two voltage levels

When I draw a specific network (2 voltage levels with a line between them), the size of the viewbox in which the diagram is drawn is not consistent and in some cases is too small, which cuts the bottom of the diagram:

network_common_methods_createLine_TB

The size of the viewbox in which the diagram is drawn should be computed so that it never cuts the diagram.

### Describe the steps

1. Load one of the IIDM files from the archive [iidm_files.zip](https://github.com/powsybl/powsybl-diagram/files/12109357/iidm_files.zip)
2. Draw the network using powsybl-diagram

You should then get one of the following files : [svg_files.zip](https://github.com/powsybl/powsybl-diagram/files/12109426/svg_files.zip)
The problematic one is the file "network_common_methods_createLine_TB.svg"
The only difference between all the IIDM files is the direction (TOP or BOTTOM) of the line on the two voltage levels.

### Environment

- powsybl-single-line-diagram-core v3.3.0

### Relevant Log Output

_No response_

### Extra Information

_No response_

tadam50:

'Fixes #536'


So-Fras (approved):
Thank you for the fix and the associated tests!

PR #535, opened 2023-07-18
Warn user when trying to use voltagelevel from another substation

tadam50:
**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Does this PR already have an issue describing the problem?**
'Fixes #534'

Issue #534, opened 2023-07-17
NullPointerException creating SubstationGraph with line


tadam50:
### Describe the current behavior

With:
- `ssb1`: a `SubstationRawBuilder`
- `ssb2`: a `SubstationRawBuilder`

If we use the `SubstationRawBuilder::createLine` method from `ssb1` to create a line between two substations `ssb1` & `ssb2`.
A `LineEdge` will be added into `ssb1` using RawGraphBuilder::buildSubstationGraph(`ssb1`);

When we want to build a substation graph from horizontal or vertical layout an exception is raised.
Indeed we are looking for a voltagelevel graph which is not in the substation to build.
This case is not managed. 

### Describe the expected behavior

No null exception must be raised.
A assertion can be raised to the user to warn him of misuse ?

### Describe the steps

In TestCase11SubstationGraph replace :
`Map<VoltageLevelRawBuilder, FeederNode> line1 = ssb2.createLine("line1", vlb1, vlsubst2, 7, 1, TOP, TOP);`
by
`Map<VoltageLevelRawBuilder, FeederNode> line1 = ssb1.createLine("line1", vlb1, vlsubst2, 7, 1, TOP, TOP);`

The following exception is raised in tests  testH():
java.lang.NullPointerException: Cannot invoke "com.powsybl.sld.model.graphs.VoltageLevelGraph.getX()" because "vlGraph2" is null

and testV():
java.lang.NullPointerException: Cannot invoke "com.powsybl.sld.model.graphs.VoltageLevelInfos.getId()" because the return value of "com.powsybl.sld.model.graphs.SubstationGraph.getVoltageLevelInfos(com.powsybl.sld.model.nodes.Node)" is null

### Relevant Log Output

testV():
com.powsybl.sld.layout.VerticalSubstationLayout.addMiddlePoints(VerticalSubstationLayout.java:174)
	at com.powsybl.sld.layout.VerticalSubstationLayout.calculatePolylineSnakeLine(VerticalSubstationLayout.java:144)
	at com.powsybl.sld.layout.AbstractLayout.manageSnakeLines(AbstractLayout.java:55)
	at com.powsybl.sld.layout.VerticalSubstationLayout.manageSnakeLines(VerticalSubstationLayout.java:70)
	at com.powsybl.sld.layout.AbstractSubstationLayout.run(AbstractSubstationLayout.java:37)
[...]

PR #543, opened 2023-07-31
Revert issue 535

tadam50:

`'Revert #535'`


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/SubstationRawBuilder.java
```diff
@@ -53,9 +54,6 @@ public Map<VoltageLevelRawBuilder, FeederNode> createLine(String id, VoltageLeve
         if (containsVoltageLevelRawBuilder(vl1) && containsVoltageLevelRawBuilder(vl2)) {
             // All VoltageLevel must be in the same Substation
```
So-Fras, line 55: Actually, lines can exist between substations (and are mainly between substations). Do we keep this condition because lines between substations will be dealt with at the zone graph level?

So-Fras (commented):
Thanks for reverting the #535 PR. I just have one question about line management, in relation with zone graph!

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/SubstationRawBuilder.java
tadam50: Indeed without these conditions an NullPointerException will be raised as described in issue #534

So-Fras (approved):
In the end, this PR is more than just a revert. We added exceptions for several cases in the SubstationRawBuilder, which are more explicit than a null pointer exception.

PR #544, opened 2023-07-31
LCC & VSC : add escaped id in metadata

tadam50:

No

Feature

In node metadata for VSC or LCC contains fields id & equipmentId.
But the field id is "ascii-encoded" & equipmentId is the HVDC line id.

Add in metadata the LCC or VSC id.

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
@@ -33,6 +33,8 @@ public class GraphMetadata {
     @JsonInclude(JsonInclude.Include.NON_NULL)
     public static class NodeMetadata {
 
+        private final String unescapedId;
+
         private final String id;
 
         private final String componentType;
@@ -51,8 +53,7 @@ public class GraphMetadata {
 
         private final List<NodeLabelMetadata> labels;
 
-        @JsonCreator
-        public NodeMetadata(@JsonProperty("id") String id,
+        public NodeMetadata(@JsonProperty("id") String escapedId,
                             @JsonProperty("vid") String vId,
                             @JsonProperty("nextVId") String nextVId,
                             @JsonProperty("componentType") String componentType,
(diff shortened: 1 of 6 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
```diff
@@ -51,8 +53,21 @@ public static class NodeMetadata {
 
         private final List<NodeLabelMetadata> labels;
 
+        public NodeMetadata(@JsonProperty("id") String escapedId,
```
So-Fras, line 56: True question: is there a need to keep this constructor? For example, for a `Node` that is not an `EquipmentNode`, the `equipmentId` is `null` and we do not have a specific method without the `equipmentId`.
Since the method is public, it would cause a breaking change to delete it.
Is deprecating it in the first place the right thing to do?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/svg/GraphMetadataTest.java
```diff
@@ -149,6 +149,7 @@ void test() throws IOException {
     @Test
     void testGraphMetadataWithLine() throws IOException {
         GraphMetadata metadata = new GraphMetadata(new LayoutParameters());
+        metadata.addNodeMetadata(new GraphMetadata.NodeMetadata("STATION_EXEMPLE", "idSTATION_95_EXEMPLE", "vid1", null, LCC_CONVERTER_STATION, false, Direction.UNDEFINED, false, null, Collections.emptyList()));
```
So-Fras, line 152: EXEMPLE is a French word ;) I would suggest EXAMPLE instead!

So-Fras (changes requested):
Thank you for working quickly and yet without very clear specifications (my bad!). 
I have a few minor comments and something (what should we do about the existing NodeMetadata constructor) I would like to discuss further with you!

So-Fras (approved):
Thank you for your work!

PR #567, opened 2023-11-21
Fix 3WT feeder info on voltage level graphs

So-Fras:

#565 

Bug fix

The feeder arrows of the 2 external legs of a three-winding transformer on a voltage level graph are displayed in the wrong direction and the feeder info values have the wrong sign.

The arrow direction and value sign are fixed.

**Other information**:
This is a quick patch. We are thinking of removing the feeder arrows on the 2 external legs of the three-winding transformer and only display the flows for the leg connected to the displayed voltage level.

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
@@ -31,8 +31,4 @@ public class FeederTwLeg extends FeederWithSides {
         }
         return super.getVoltageLevelInfos();
     }
-
-    public VoltageLevelInfos getOwnVoltageLevelInfos() {
-        return ownVoltageLevelInfos;
-    }
 }
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
@@ -15,11 +15,8 @@ import com.powsybl.sld.model.nodes.NodeSide;
  */
 public class FeederTwLeg extends FeederWithSides {
 
-    VoltageLevelInfos ownVoltageLevelInfos;
-
     public FeederTwLeg(FeederType feederType, NodeSide side, VoltageLevelInfos myVoltageLevelInfos, VoltageLevelInfos otherSideVoltageLevelInfos) {
         super(feederType, side, myVoltageLevelInfos, otherSideVoltageLevelInfos);
-        this.ownVoltageLevelInfos = myVoltageLevelInfos;
     }
 
     @Override
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
@@ -15,8 +15,11 @@ import com.powsybl.sld.model.nodes.NodeSide;
  */
 public class FeederTwLeg extends FeederWithSides {
 
+    private final VoltageLevelInfos ownVoltageLevelInfos;
+
     public FeederTwLeg(FeederType feederType, NodeSide side, VoltageLevelInfos myVoltageLevelInfos, VoltageLevelInfos otherSideVoltageLevelInfos) {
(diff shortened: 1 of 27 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
```diff
@@ -15,8 +15,11 @@
  */
 public class FeederTwLeg extends FeederWithSides {
 
+    VoltageLevelInfos ownVoltageLevelInfos;
```
flo-dup, line 18:
```suggestion
    private final VoltageLevelInfos ownVoltageLevelInfos;
```
Parameters should usually be private and accessed through getters. When they don't need to change they should also be marked final

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -859,6 +860,13 @@ protected void insertFeederInfos(String prefixId,
 
         double shiftFeederInfo = 0;
         for (FeederInfo feederInfo : labelProvider.getFeederInfos(feederNode)) {
+            String voltageLevelId = graph.getVoltageLevelInfos().getId();
+            if (feederNode.getFeeder() instanceof FeederTwLeg feederTwLeg) {
+                String feederTwLegVoltageLevelId = feederTwLeg.getVoltageLevelInfos().getId();
+                if (!voltageLevelId.equals(feederTwLegVoltageLevelId) && feederInfo instanceof DirectionalFeederInfo directionalFeederInfo) {
+                    directionalFeederInfo.reverseArrowDirectionAndValue();
```
flo-dup, line 867: I think this reverse should be done within the label provider. If a user of powsybl-diagram was using its own labelProvider you're reversing his results.

It would also be better to do it at the FeederInfo construction, not afterwards.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DirectionalFeederInfo.java
```diff
@@ -19,9 +19,9 @@
  */
 public class DirectionalFeederInfo extends AbstractFeederInfo {
 
-    private final LabelProvider.LabelDirection arrowDirection;
+    private LabelProvider.LabelDirection arrowDirection;
```
flo-dup, line 22: If possible we should keep those final.

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/feeders/FeederTwLeg.java
So-Fras: I forgot to delete this attribute actually. I do not use it at the moment.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DirectionalFeederInfo.java
So-Fras: Yes, I agree. For that, I must perform the reverse within the label provider.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
So-Fras: Agreed, I will look for a way to do that without breaking the API.

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/Test3WTFeederInfoArrow.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author Sophie Frasnedo {@literal <sophie.frasnedo at rte-france.com>}
+ */
+class Test3WTFeederInfoArrow extends AbstractTestCaseIidm {
+
+    @BeforeEach
+    public void setUp() {
+        network = ThreeWindingsTransformerNetworkFactory.create();
+        network.getThreeWindingsTransformer("3WT").getLeg1().getTerminal().setP(1400.0).setQ(400.0);
+        network.getThreeWindingsTransformer("3WT").getLeg2().getTerminal().setP(1400.0).setQ(400.0);
+        network.getThreeWindingsTransformer("3WT").getLeg3().getTerminal().setP(1400.0).setQ(400.0);
```
flo-dup, line 29: Could you just put values whose sum is more or less equals to 0?

flo-dup (approved):

PR #568, opened 2023-11-21
Update CI actions due to node.js v12 deprecation

So-Fras:

No but similar to this PR on powsybl-core : https://github.com/powsybl/powsybl-core/pull/2766

CI was using [setup-java](https://github.com/actions/setup-java/) v2 and [checkout](https://github.com/actions/checkout/) v2.3.4

CI is now using setup-java v3 and checkout v4


rolnico (approved):

.github/workflows/maven.yml
```diff
@@ -18,12 +18,12 @@ jobs:
 
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v4
```
flo-dup, line 21: you need to put the hash corresponding to the release, as the tag might be changed by an attacker

flo-dup: You can look at powsybl/powsybl-core#2821 for an example

So-Fras: Ok, thanks for pointing it out. I took more recent version too, following the example in core: 
- 4.1.1 for actions/checkout instead of 4
- 4.0.0 for actions/setup-java instead of 3

PR #580, opened 2024-01-26
Fix macOS compilation issues

tadam50:

No

Bug fix

MacOS build is not always sucessful on the same test failure (TestCase11SubstationGraph::testWithHvdcLines)

Result of compareMetadata is not a simple boolean but return expected and actual metadata.
This, in order to ba able to use assertEquals instead of assertTrue to get more information at failure

- [ ] Yes
- [x] No

**Other information**:


single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/AbstractTestCase.java
```diff
@@ -135,7 +136,7 @@ public String toSVG(Graph graph, String filename, ComponentLibrary componentLibr
         }
     }
 
-    public boolean compareMetadata(VoltageLevelGraph graph, String refMetadataName, VoltageLevelLayoutFactory voltageLevelLayoutFactory, ComponentLibrary componentLibrary, LayoutParameters layoutParameters, SvgParameters svgParameters, LabelProvider labelProvider, StyleProvider styleProvider) {
+    public JsonNode[] compareMetadata(VoltageLevelGraph graph, String refMetadataName, VoltageLevelLayoutFactory voltageLevelLayoutFactory, ComponentLibrary componentLibrary, LayoutParameters layoutParameters, SvgParameters svgParameters, LabelProvider labelProvider, StyleProvider styleProvider) {
```
flo-dup, line 139: It's not a comparison which is done in there, so the name is misleading

flo-dup (changes requested):
Why not doing the same test process than with svg or graph json? That is, in `AbstractTestCase`, 
```java
public abstract String toMetadata(Graph g, String filename);
public String toMetadata(Graph graph, String refMetadataName, LabelProvider labelProvider, StyleProvider styleProvider) {
  ...
}

```
which is then called in the unit tests by
```java
assertEquals(toString("/metadata.json"), toMetadata(graph, "/metadata.json)
```

This should allow us to identify why it fails on macOS. Maybe we'll do the `JsonNode` comparison at the end, but here as the test issue is not completely identified we can't be sure that it's a good solution.

PR #591, opened 2024-02-21
Add scorecard to powsybl-diagram

So-Fras:

No

This PR adds another GitHub action (Scorecard) to the powsybl-diagram workflow

- [ ] Yes
- [x] No


So-Fras:
I tried to create a fine-grained PAT as described [here](https://github.com/ossf/scorecard-action?tab=readme-ov-file#authentication-with-fine-grained-pat-optional). 
It seems to me that we may need it because we use branch protection rules.
However, I cannot go through the "select the powsybl-diagram repository" step (I think it is because I am not an owner of the repository).

So-Fras:
This PR was a test. In the end, we decided not to add the scorecard GitHub action to the powsybl-diagram workflow (and other repositories' workflows).

PR #593, opened 2024-03-05
Spread voltageLevels with missing position information

So-Fras:

No

Feature

When a voltageLevel has no fixed position (the position information may be missing), it is displayed in (0, 0) when using the BasicFixedLayoutFactory.

It is possible to pass a LayoutFactory (for example a BasicForceLayoutFactory) as an argument to the BasicFixedLayoutFactory create(...) fucntion to spread voltageLevels that have missing position information.

- [ ] Yes
- [x] No


network-area-diagram/src/main/java/com/powsybl/nad/layout/ImprovedForceLayoutFactory.java
flo-dup, line 1: The layout is not really improved, we should rather name it `FixedPositionsLayoutFactory`.

Shouldn't you merge this factory with `BasicFixedLayoutFactory`?

network-area-diagram/src/main/java/com/powsybl/nad/layout/ImprovedForceLayoutFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    private final Map<String, Point> fixedPositions;
+
+    public ImprovedForceLayoutFactory(Map<String, Point> fixedPositions) {
+        Objects.requireNonNull(fixedPositions);
+        this.fixedPositions = fixedPositions;
+    }
+
+    @Override
+    public Layout create() {
+        AbstractLayout layout = new BasicForceLayout();
```
flo-dup, line 29: Here's `BasicForceLayout` but in fact we may want fixed positions on any `Layout` algorithms, so it would be nicer to have a factory as constructor argument.

If that factory is null, the fallback could be the `BasicFixedLayout`

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicFixedLayoutFactory.java
```diff
@@ -29,4 +29,15 @@ public Layout create() {
         layout.setFixedNodePositions(fixedPositions);
         return layout;
     }
+
+    public Layout create(LayoutFactory layoutFactory) {
```
flo-dup, line 33: We'd like to have this part of code inside `public Layout create()` instead so that it gets called in `NetworkAreaDiagram::draw`.

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicFixedLayoutFactory.java
```diff
@@ -29,4 +29,15 @@ public Layout create() {
...
+        if (layoutFactory == null) {
+            return create();
+        } else {
+            Layout layout = layoutFactory.create();
+            layout.setInitialNodePositions(fixedPositions);
+            layout.setNodesWithFixedPosition(fixedPositions.keySet());
```
flo-dup, line 39: You see here that the current `Layout` API is not very convenient, we should modify it.

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicFixedLayoutFactory.java
So-Fras: True! I created an issue for that: https://github.com/powsybl/powsybl-diagram/issues/601

flo-dup (approved):

PR #595, opened 2024-03-18
Fix one leg intern cells

flo-dup:

No

Bug fix

One-leg intern cells, which are intern cells which are displayed as a single vertical line, forces the corresponding busbars to be parallel, which used to lead to unexpected behaviour (mixing components on one of the two bars). Since PR #573, an exception is thrown.

Example of one-leg intern cell from unit tests:
one-leg

- Wannabe one-leg intern cells are detected when organizing blocks like previously but the corresponding shape is now `MAYBE_ONE_LEG`
- Then, when building the LegBusSets, the maybe-one-leg cell is
   - either integrated in an existing LegBusSet if all its busNodes are contained in that one, its shape is then confirmed `ONE_LEG`
   - or split into two LegBlocks based on the LegBusSets it is attached to and its shape is either marked `UNDEFINED` or `MAYBE_FLAT`
   - or split into two LegBlocks based on the sectionIndex of their BusNode and its shape is either marked `UNDEFINED` or `MAYBE_FLAT`
   - or marked as `ONE_LEG`
- Later, when creating subsections, if we detect that the maybe-one-leg cell which was split into two parts is fully included in a subsection, we finally set its shape to `ONE_LEG`

Note that when splitting in two parts the subBlocks of the internCell, we favour the splitting based on LegBusSets as they can deal with complex subsections which contain more than one sectionIndex (e.g., two parallel busbars with different lengths).

- [ ] Yes
- [x] No


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/LegBusSet.java
```diff
@@ -232,15 +231,59 @@ public static void pushLBS(List<LegBusSet> legBusSets, LegBusSet legBusSet) {
...
+    private static void replaceByMultilegOrSetOneLeg(InternCell internCell, List<LegBusSet> attachedLegBusSets) {
+        // the one leg intern cell cannot be absorbed by any existing legBusSet
+        // we consider that a one leg intern cell should not force the corresponding busNodes
+        // to be parallel, hence we try to replace that one leg by a multileg
+        LegBlock oneLeg = internCell.getSideToLeg(Side.UNDEFINED);
+        if (oneLeg instanceof LegParallelBlock legParallelBlock) {
+            List<LegPrimaryBlock> subBlocks = legParallelBlock.getSubBlocks();
+            if (subBlocks.size() == 2) {
+                internCell.replaceOneLegByMultiLeg(subBlocks.get(0), subBlocks.get(1));
+            } else {
```
So-Fras, line 271: This else statement is never used (neither in unit tests nor in other tests I ran with French substations).
Shall we remove it?

So-Fras (changes requested):

PR #615, opened 2024-04-30
Add edge type to metadata in network area diagram SVG

massimo-ferraro:
- [X] Tests for the changes have been modified (for bug fixes / features)

no

feature

Information about the edge type has been added to the metadata in the network area diagram SVG

- [ ] Yes
- [X] No

### DIFF
--- a/network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
+++ b/network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
@@ -899,20 +899,21 @@ public class SvgWriter {
                 busNode.getEquipmentId(),
                 String.valueOf(busNode.getNbNeighbouringBusNodes()),
                 String.valueOf(busNode.getIndex()),
-                getPrefixedId(vlNode.getDiagramId())
-        )));
+                getPrefixedId(vlNode.getDiagramId()))));
         graph.getNodesStream().forEach(node -> metadata.addNode(getPrefixedId(node.getDiagramId()), node.getEquipmentId(),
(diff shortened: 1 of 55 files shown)


network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -899,20 +899,21 @@ private void addMetadata(Graph graph, XMLStreamWriter writer) throws XMLStreamEx
...
-                getPrefixedId(graph.getBusGraphNode2(edge).getDiagramId())));
+                getPrefixedId(graph.getBusGraphNode2(edge).getDiagramId()),
+                edge.getType()));
         graph.getThreeWtEdgesStream().forEach(edge -> metadata.addEdge(getPrefixedId(edge.getDiagramId()), edge.getEquipmentId(),
                 getPrefixedId(graph.getNode1(edge).getDiagramId()),
                 getPrefixedId(graph.getNode2(edge).getDiagramId()),
                 getPrefixedId(graph.getBusGraphNode1(edge).getDiagramId()),
-                getPrefixedId(graph.getBusGraphNode2(edge).getDiagramId())));
+                getPrefixedId(graph.getBusGraphNode2(edge).getDiagramId()),
+                "ThreeWtEdge"));
```
flo-dup, line 916: It would be cleaner if `ThreeWtEdge` holds that string type.

In fact we should move the type attribute which is in `BranchEdge` in `AbstractEdge`, and add `getType()` method in the `Edge` empty interface. Could you do such a change in this PR?

PR #616, opened 2024-05-03
Do not compute forces, velocity and position for fixed points

flo-dup:

No

"Bug" fix

Forces and velocity are computed for fixed points

Forces and velocity are **not** computed for fixed points

- [ ] Yes
- [x] No


diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
```diff
@@ -160,18 +161,26 @@ private void initializePoints() {
         setCenter(initialPointsCenter.orElse(new Vector(0, 0)));
 
         for (V vertex : graph.vertexSet()) {
-            Point point = new Point(
-                    center.getX() + scale * (random.nextDouble() - 0.5),
-                    center.getY() + scale * (random.nextDouble() - 0.5)
-            );
-            points.put(vertex, initialPoints.getOrDefault(vertex, point));
+            if (!fixedNodes.contains(vertex)) {
+                points.put(vertex, new Point(
```
So-Fras, line 165: Here, the `initialPoints` are not used anymore. It is also weird that no test has failed after this modification (we may have not used `initialPoints` in tests)

flo-dup: Indeed, I forgot the case of non-fixed initial points!

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
```diff
@@ -302,15 +314,10 @@ private void updateVelocity() {
     }
 
     private void updatePosition() {
-        // TODO do not compute forces or update velocities for fixed nodes
         // We have computed forces and velocities for all nodes, even for the fixed ones
```
So-Fras, line 317: `// We have computed forces and velocities for all nodes, even for the fixed ones`
`// We can optimize calculations by ignoring fixed nodes in those calculations`

Are those comments up-to-date? Is velocity still computed for fixed points and is there still a possibility for optimization?

So-Fras (approved):

PR #639, opened 2024-08-06
Add layout parameters to metadata in network area diagram SVG

massimo-ferraro:
- [X] Tests for the changes have been modified (for bug fixes / features)

no

feature

Layout parameters have been added to the metadata in the network area diagram SVG

- [ ] Yes
- [X] No


network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -944,6 +948,9 @@ private void addMetadata(Graph graph, XMLStreamWriter writer) throws XMLStreamEx
                                   String.valueOf(svgParameters.getPowerValuePrecision()), String.valueOf(svgParameters.getAngleValuePrecision()),
                                   String.valueOf(svgParameters.getCurrentValuePrecision()), svgParameters.getEdgeInfoDisplayed().name(),
                                   getFormattedValue(svgParameters.getPstArrowHeadSize()), svgParameters.getUndefinedValueSymbol());
+        metadata.addLayoutParameters(String.valueOf(layoutParameters.isTextNodesForceLayout()), getFormattedValue(layoutParameters.getSpringRepulsionFactorForceLayout()),
```
flo-dup, line 951: using `getFormattedValue` makes sense when you're writing metadata about x/y coordinates which are also in the svg, as in the svg the coordinates are written with a fixed precision thanks to this method. But I don't think you should use it for all other values, especially the one which might need more precision (spring repulsion factor, fixed scale, the angles in radians loopEdgesAperture and edgesForkAperture, ...)

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
massimo-ferraro: Put original values instead of rounded formatted values for parameters in metadata

flo-dup (approved):

Issue #590, opened 2024-02-02
StyleProvider to highlight overloaded lines in SLD


So-Fras:
### Describe the current behavior

In network-area-diagram, branches and tie lines are set to blink when they are overloaded (i.e. when the current permanent limit is violated).

In single-line-diagram, such a feature does not exist.

### Describe the expected behavior

Tie lines, transformer legs, lines, dangling lines that whose current permanent limit is violated should be highlighted with a blinking design supported by a new StyleProvider.

### Describe the motivation

This piece of information is also useful in single-line-diagram

PR #643, opened 2024-09-03
[SLD] add sld-overload class to equipment label

jamal-khey:

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
@@ -336,13 +336,23 @@ public class NetworkGraphBuilder implements GraphBuilder {
                         voltageLevelInfosBySide.get(NodeSide.TWO),
                         voltageLevelInfosBySide.get(NodeSide.THREE));
 
+                if (transformer.isOverloaded()) {
+                    //control the container
+                    middleNode.setInOverload(true);
+                    firstOtherLegNode.setInOverload(true);
+                    secondOtherLegNode.setInOverload(true);
+
+                }
                 add3wtFeeder(middleNode, firstOtherLegNode, secondOtherLegNode, transformer.getTerminal(side));
             } else {
                 // in substation diagram, we only represent the leg node within the voltage level (3wt node will be on the snake line)
                 String id = transformer.getId() + "_" + side.name();
                 FeederNode legNode = NodeFactory.createFeeder3WTLegNodeForSubstationDiagram(graph, id, transformer.getNameOrId(), transformer.getId(),
                         NodeSide.valueOf(side.name()));
-
+                if (transformer.isOverloaded()) {
+                    //control the container
+                    legNode.setInOverload(true);
+                }
                 addTerminalNode(legNode, transformer.getTerminal(side));
             }
         }
@@ -403,13 +413,20 @@ public class NetworkGraphBuilder implements GraphBuilder {
 
         @Override
         public void visitTwoWindingsTransformer(TwoWindingsTransformer transformer, TwoSides side) {
-            Node transformerNode = createFeeder2wtNode(graph, transformer, side);
+            FeederNode transformerNode = createFeeder2wtNode(graph, transformer, side);
+            if (transformer.isOverloaded()) {
+                transformerNode.setInOverload(true);
+            }
             addTerminalNode(transformerNode, transformer.getTerminal(side));
         }
 
         @Override
         public void visitLine(Line line, TwoSides side) {
-            addTerminalNode(createFeederLineNode(graph, line, side), line.getTerminal(side));
+            FeederNode feederNode = createFeederLineNode(graph, line, side);
+            if (line.isOverloaded()) {
+                feederNode.setInOverload(true);
+            }
+            addTerminalNode(feederNode, line.getTerminal(side));
         }
 
         private static VoltageLevelInfos createVoltageLevelInfos(Terminal terminal) {
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/Node.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/Node.java
@@ -76,6 +76,10 @@ public interface Node {
 
     Direction getDirection();
 
+    boolean isInOverload();
+
+    void setInOverload(boolean isInOverload);
+
     void setDirection(Direction direction);
 
     Orientation getOrientation();
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/AbstractNode.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/AbstractNode.java
@@ -46,6 +46,8 @@ public abstract class AbstractNode implements Node {
 
     private Orientation orientation;
 
+    private boolean isInOverload;
+
     /**
      * Constructor
      */
@@ -54,6 +56,7 @@ public abstract class AbstractNode implements Node {
         this.id = Objects.requireNonNull(id);
         this.componentType = Objects.requireNonNull(componentType);
         this.fictitious = fictitious;
+        this.isInOverload = false;
         setOrientation(defaultOrientation());
     }
 
@@ -190,6 +193,16 @@ public abstract class AbstractNode implements Node {
     @Override
     public void setOrientation(Orientation orientation) {
         this.orientation = Objects.requireNonNullElse(orientation, defaultOrientation());
+    }
+
+    @Override
+    public boolean isInOverload() {
+        return isInOverload;
+    }
+
+    @Override
+    public void setInOverload(boolean isInOverload) {
+        this.isInOverload = isInOverload;
     }
 
     @Override
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
@@ -509,7 +509,8 @@ public class DefaultSVGWriter implements SVGWriter {
         for (LabelProvider.NodeLabel nodeLabel : nodeLabels) {
 
             LabelPosition labelPosition = nodeLabel.getPosition();
-            Element label = createLabelElement(nodeLabel.getLabel(), labelPosition.getdX(), labelPosition.getdY(), labelPosition.getShiftAngle(), g);
+            Element label = createLabelElement(nodeLabel.getLabel(), labelPosition.getdX(), labelPosition.getdY(), labelPosition.getShiftAngle(), g, node.isInOverload());
+
             String svgId = getNodeLabelId(prefixId, node, labelPosition);
             label.setAttribute("id", svgId);
             if (labelPosition.isCentered()) {
@@ -541,7 +542,7 @@ public class DefaultSVGWriter implements SVGWriter {
         double yPos = graph.getY() - 20.;
 
         String graphName = svgParameters.isUseName() ? graph.getVoltageLevelInfos().getName() : graph.getVoltageLevelInfos().getId();
-        Element label = createLabelElement(graphName, graph.getX(), yPos, 0, gLabel);
+        Element label = createLabelElement(graphName, graph.getX(), yPos, 0, gLabel, false);
         label.setAttribute(CLASS, StyleClassConstants.GRAPH_LABEL_STYLE_CLASS);
         gLabel.appendChild(label);
         root.appendChild(gLabel);
@@ -578,7 +579,7 @@ public class DefaultSVGWriter implements SVGWriter {
     /*
      * Create a label text element at the given position
      */
-    protected Element createLabelElement(String str, double xShift, double yShift, int shiftAngle, Element g) {
+    protected Element createLabelElement(String str, double xShift, double yShift, int shiftAngle, Element g, boolean inOverload) {
         Element label = g.getOwnerDocument().createElement("text");
         label.setAttribute("x", String.valueOf(xShift));
         label.setAttribute("y", String.valueOf(yShift));
@@ -586,6 +587,10 @@ public class DefaultSVGWriter implements SVGWriter {
             label.setAttribute(TRANSFORM, ROTATE + "(" + shiftAngle + "," + 0 + "," + 0 + ")");
         }
         label.setAttribute(CLASS, StyleClassConstants.LABEL_STYLE_CLASS);
+        if (inOverload) {
+            String labelClass = label.getAttribute(CLASS); // add the overload class to the label
+            label.setAttribute(CLASS, labelClass + " " + StyleClassConstants.OVERLOAD);
+        }
         Text text = g.getOwnerDocument().createTextNode(str);
         label.appendChild(text);
         return label;
@@ -906,13 +911,13 @@ public class DefaultSVGWriter implements SVGWriter {
 
         // we draw the right label only if present
         feederInfo.getRightLabel().ifPresent(s -> {
-            Element labelRight = createLabelElement(s, shX, shY, 0, g);
+            Element labelRight = createLabelElement(s, shX, shY, 0, g, false);
             g.appendChild(labelRight);
         });
 
         // we draw the left label only if present
         feederInfo.getLeftLabel().ifPresent(s -> {
-            Element labelLeft = createLabelElement(s, -LABEL_OFFSET, shY, 0, g);
+            Element labelLeft = createLabelElement(s, -LABEL_OFFSET, shY, 0, g, false);
             labelLeft.setAttribute(STYLE, "text-anchor:end");
             g.appendChild(labelLeft);
         });
@@ -960,13 +965,13 @@ public class DefaultSVGWriter implements SVGWriter {
 
         // We draw the bottom label only if present
         busInfo.getBottomLabel().ifPresent(s -> {
-            Element labelBottom = createLabelElement(s, 0, shY, 0, g);
+            Element labelBottom = createLabelElement(s, 0, shY, 0, g, false);
             g.appendChild(labelBottom);
         });
 
         // We draw the top label only if present
         busInfo.getTopLabel().ifPresent(s -> {
-            Element labelTop = createLabelElement(s, 0, -LABEL_OFFSET, 0, g);
+            Element labelTop = createLabelElement(s, 0, -LABEL_OFFSET, 0, g, false);
             g.appendChild(labelTop);
         });
         root.appendChild(g);
@@ -996,6 +1001,11 @@ public class DefaultSVGWriter implements SVGWriter {
         for (Edge edge : edges) {
             String wireId = getWireId(prefixId, voltageLevelId, edge);
 
+            if (edge instanceof BranchEdge) {
+                String id = ((BranchEdge) edge).getId();
+                System.out.println("id= " + id);
+            }
+
             List<Point> pol = new ArrayList<>();
             if (!edge.isZeroLength()) {
                 // Determine points of the polyline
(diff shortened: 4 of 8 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/AbstractVoltageStyleProvider.java
```diff
@@ -49,6 +49,9 @@ public List<String> getEdgeStyles(Graph graph, Edge edge) {
                 .map(Optional::get)
                 .collect(Collectors.toList());
         list.add(0, WIRE_STYLE_CLASS);
+        if (edge.isOverloaded()) {
+            list.add(StyleClassConstants.OVERLOAD_STYLE_CLASS);
+        }
```
EstherDarkish, line 54: Can you clean the WIRE_STYLE_CLASS just above your change :
`list.add(0, WIRE_STYLE_CLASS);` => `list.add(0, StyleClassConstants.WIRE_STYLE_CLASS);`
and remove the static import line 24

flo-dup:
> Shouldn't we add the style in the .css?

Indeed, could you add the css from gridsuite/gridstudy-app#2244?


> What kind of equipment could be overloaded? Here I am under the impression that a connectivityNode could be "overloaded". Does it make sense?

Indeed, I think overloaded is now everywhere in the graph, but we tried to avoid to populate the graph with data which is not needed for the diagram structure. The way we handle this is to look in the network when requesting the classes on a particular node or a particular edge, like in `HighlightLineStateStyleProvider`


> I am not sure that we should modify the AbstractStyleProvider. We could do the same as what we did with the `HighlightLineStateStyleProvider`

Indeed, we'd rather do it like that

flo-dup:
About the rendering, if we want to keep the same rendering as the connect / disconnect styles from `HighlightLineStateStyleProvider`, we should have only the last edge blinking, not the whole cell like in the unit tests. And if it's a transformer which is overloaded, shouldn't it be only the transformer which is blinking?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
```diff
@@ -68,6 +68,14 @@ private static boolean isNotInternalToSubstation(Branch<?> branch) {
         return !isInternalToSubstation(branch);
     }
 
+    public static boolean isOverLoaded(Branch<?> branch) {
```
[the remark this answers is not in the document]
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
```diff
@@ -403,13 +418,20 @@ public void visitStaticVarCompensator(StaticVarCompensator svc) {
 
         @Override
         public void visitTwoWindingsTransformer(TwoWindingsTransformer transformer, TwoSides side) {
-            Node transformerNode = createFeeder2wtNode(graph, transformer, side);
+            FeederNode transformerNode = createFeeder2wtNode(graph, transformer, side);
+            if (isOverLoaded(transformer)) {
```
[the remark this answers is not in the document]
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/builders/NetworkGraphBuilder.java
```diff
@@ -403,13 +418,20 @@ public void visitStaticVarCompensator(StaticVarCompensator svc) {
...
+            }
             addTerminalNode(transformerNode, transformer.getTerminal(side));
         }
 
         @Override
         public void visitLine(Line line, TwoSides side) {
-            addTerminalNode(createFeederLineNode(graph, line, side), line.getTerminal(side));
+            FeederNode feederNode = createFeederLineNode(graph, line, side);
+            if (isOverLoaded(line)) {
+                feederNode.setInOverload(true);
```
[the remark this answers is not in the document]
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/nodes/Node.java
```diff
@@ -76,6 +76,10 @@ enum NodeType {
 
     Direction getDirection();
 
+    boolean isOverloaded();
+
+    void setInOverload(boolean isInOverload);
```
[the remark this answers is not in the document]
jamal-khey: done

jamal-khey:
> > What kind of equipment could be overloaded? Here I am under the impression that a connectivityNode could be "overloaded". Does it make sense?
>I want to be able to make the icons representing transformers blink as well. These transformers are represented by nodes of the type ConnectivityNode.

So-Fras:
> Indeed, I think overloaded is now everywhere in the graph, but we tried to avoid to populate the graph with data which is not needed for the diagram structure. The way we handle this is to look in the network when requesting the classes on a particular node or a particular edge, like in `HighlightLineStateStyleProvider`


I reread the PR and realized that we did not fully address @flo-dup 's comment. What he means is that, if possible, we do not want to add the "isOverloaded" (or "isLimitExceeded") kind of information to the graph because it is not necessary to build it. The graph is about `Node` and `Edge` and we try not to include "power network" kind of information in it.

When we had to do something similar to your PR in the past, we proceeded as in the `HighlightLineStateStyleProvider`. We linked the `Node` or `Edge` to equipment of the network and then we assessed the "state" of the `Node` or `Edge`. That is why the `HighlightLineStateLineStyleProvider`'s constructor takes a `Network` as a parameter.

Could you try to do the same here? Else in the `HighlightLineStateStyleProvider` (but the name of the class should be changed) or in another specific StyleProvider (maybe better).

jamal-khey:
> > Indeed, I think overloaded is now everywhere in the graph, but we tried to avoid to populate the graph with data which is not needed for the diagram structure. The way we handle this is to look in the network when requesting the classes on a particular node or a particular edge, like in `HighlightLineStateStyleProvider`
> 
> I reread the PR and realized that we did not fully address @flo-dup 's comment. What he means is that, if possible, we do not want to add the "isOverloaded" (or "isLimitExceeded") kind of information to the graph because it is not necessary to build it. The graph is about `Node` and `Edge` and we try not to include "power network" kind of information in it.

Yes you are right , i miss understood the comment.

> When we had to do something similar to your PR in the past, we proceeded as in the `HighlightLineStateStyleProvider`. We linked the `Node` or `Edge` to equipment of the network and then we assessed the "state" of the `Node` or `Edge`. That is why the `HighlightLineStateLineStyleProvider`'s constructor takes a `Network` as a parameter.
> 
> Could you try to do the same here? Else in the `HighlightLineStateStyleProvider` (but the name of the class should be changed) or in another specific StyleProvider (maybe better).

I have made the necessary changes to align with this approach, similar to the `HighlightLineStateStyleProvider` , i added `LimitHighlightStyleProvider`

single-line-diagram/single-line-diagram-core/src/main/resources/animations.css
```diff
@@ -31,3 +31,57 @@
 .sld-arrow-animation-low-speed { --sld-arrow-animation-parameters: 3s infinite linear }
 .sld-arrow-animation-average-speed { --sld-arrow-animation-parameters: 2s infinite linear }
 .sld-arrow-animation-high-speed { --sld-arrow-animation-parameters: 1s infinite linear }
+
+:root {
+  --sld-overload-color: #ffeb3b; /* Yellow for overload */
+  --sld-overvoltage-color: #5a0101; /* Red for overvoltage */
+  --sld-undervoltage-color: #0000ff; /* Blue for undervoltage */
+}
```
flo-dup, line 39: They're constant and they're used only once each, so you can use directly the value inside the animations.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+import com.powsybl.sld.model.nodes.feeders.FeederWithSides;
+import com.powsybl.sld.svg.styles.EmptyStyleProvider;
+import java.util.*;
+
+import static com.powsybl.sld.svg.styles.StyleClassConstants.*;
+
+/**
+ * @author Jamal KHEYYAD {@literal <jamal.kheyyad at rte-international.com>}
+ */
+public class LimitHighlightStyleProvider extends EmptyStyleProvider {
```
flo-dup, line 17: `OverloadHighlightStyleProvider`?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+    Network network;
```
flo-dup, line 18: this can be private final

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+    public List<String> getEdgeStyles(Graph graph, Edge edge) {
+        Optional<String> overloadStyle = getOverloadStyle(edge);
+        return overloadStyle.map(Collections::singletonList).orElse(Collections.emptyList());
+    }
+
+    private Optional<String> getOverloadStyle(Edge edge) {
+        List<Node> nodes = edge.getNodes();
+        for (Node node : nodes) {
+            if (node instanceof FeederNode feederNode) {
+                return getHighlightFeederStateStyle(feederNode);
```
flo-dup, line 34: no need for an intermediate method
```java
return isOverloaded(feederNode) ? Optional.of(OVERLOAD_STYLE_CLASS) : Optional.empty();
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+            } else if (node instanceof ConnectivityNode connectivityNode && hasOverloadedAdjacentNode(connectivityNode)) {
```
flo-dup, line 35: Do you really need to extend the animation on the next edge? it gives a non-uniform length on the diagram, as it depends on how many switches there are on the cell. It does look a bit clumsy with those different lengths, and it is not consistent with the connection state style provider

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+                return Optional.of(OVERLOAD_STYLE_CLASS);
+            }
+        }
+        return Optional.empty();
+    }
+
+    boolean hasOverloadedAdjacentNode(ConnectivityNode connectivityNode) {
+        return connectivityNode.getAdjacentNodes().stream().anyMatch(node -> {
```
flo-dup, line 43: Use filter on the stream instead of that code block
```java
return connectivityNode.getAdjacentNodes().stream()
    .filter(FeederNode.class::isInstance)
    .anyMatch(node -> isOverloaded((FeederNode) feederNode));
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+            return Optional.of(OVERLOAD_STYLE_CLASS);
+        }
+        return Optional.empty();
+    }
+
+    private boolean isOverloaded(FeederNode n) {
+        if (!(n.getFeeder() instanceof FeederWithSides)) {
+            return false;
+        }
+        Connectable<?> connectable = network.getConnectable(n.getEquipmentId());
```
flo-dup, line 62: I think that we should avoid this method (which calls `getIdentifiable()`!) to be faster with powsybl-network-store. The `FeederNode::getFeederType` is unfortunately not precise enough :unamused: We need to fix this in powsybl-diagram (later).

Calling `network::getBranch` would be a bit better it seems if I look into the network-store implementation. And then calling `network::getThreeWindingsTransformer` if not found.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
...
+    public List<String> getNodeStyles(VoltageLevelGraph graph, Node node, ComponentLibrary componentLibrary, boolean showInternalNodes) {
+        if (!(node instanceof BusNode busNode)) {
+            return Collections.emptyList();
+        }
+        List<String> styles = new ArrayList<>();
+        BusbarSection busbarSection = this.network.getBusbarSection(busNode.getEquipmentId());
+        if (busbarSection != null) {
+
+            if (busbarSection.getV() > busbarSection.getTerminal().getVoltageLevel().getHighVoltageLimit()) {
+                styles.add(VL_OVERVOLTAGE_CLASS);
```
flo-dup, line 80: Why not simply `return List.of(VL_OVERVOLTAGE_CLASS)`?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
```diff
@@ -0,0 +1,89 @@
+package com.powsybl.sld.svg.styles.iidm;
```
flo-dup, line 1: Copyright missing

single-line-diagram/single-line-diagram-core/src/main/resources/animations.css
```diff
@@ -31,3 +31,57 @@
 .sld-arrow-animation-low-speed { --sld-arrow-animation-parameters: 3s infinite linear }
 .sld-arrow-animation-average-speed { --sld-arrow-animation-parameters: 2s infinite linear }
 .sld-arrow-animation-high-speed { --sld-arrow-animation-parameters: 1s infinite linear }
+
+:root {
+  --sld-overload-color: #ffeb3b; /* Yellow for overload */
+  --sld-overvoltage-color: #5a0101; /* Red for overvoltage */
```
flo-dup, line 37: this is very brown, and very close to the "fictitious" color. Besides, we should try to have the same colour as on network-area-diagram side.

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/resources/animations.css
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: first-edge
next-edge

jamal-khey: The product owner requested to extend the highlight to the switch. but this can be discussed

flo-dup: It doesn't always extend the highlight to the switch as you're not propagating the style, you're just stopping at next edge(s):
- if there is an edge linking it to the next cell (shunt) it will blink also
- if there's no switch in this cell the blinking will stop before the busbar (because there's an internal fictitious node there)

I think we should keep it simple and similar to the connectivity state to have the feature ready for the release, and discuss back with users (or UX designer?) to see what would make more sense - and remains "good-looking" for next time.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
flo-dup: woops sorry I wanted to remove that comment. The previous name was more consistent, I forgot about the busbars when I started to review

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/OverloadHighlightStyleProvider.java
```diff
@@ -0,0 +1,75 @@
...
+        return Optional.empty();
+    }
+
+    private boolean isOverloaded(FeederNode n) {
+        if (!(n.getFeeder() instanceof FeederWithSides)) {
+            return false;
+        }
+        Branch<?> branch = network.getBranch(n.getEquipmentId());
+        if (branch != null && branch.isOverloaded()) {
+            return true;
```
flo-dup, line 52: if the branch is not null and not overloaded you don't want to look for a three-winding transformer, so you'd rather write
```suggestion
        if (branch != null) {
            return branch.isOverloaded();
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/LimitHighlightStyleProvider.java
jamal-khey: done. i did the change to keep it simple

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/OverloadHighlightStyleProvider.java
jamal-khey: done

single-line-diagram/single-line-diagram-core/src/main/resources/animations.css
jamal-khey: agree , i changed it

flo-dup (approved):

So-Fras:
This PR fixes the #590 issue.

So-Fras (approved):

PR #656, opened 2024-10-23
Set layout node positions getting them from JSON metadata in network area diagram

massimo-ferraro:

no

feature

In network area diagram initial positions of nodes and text nodes for a layout can be set providing them in maps

Initial positions of nodes and text nodes for a layout can be set reading them from a JSON diagram metadata file

- [ ] Yes
- [X] No


flo-dup (approved):
I did some basic refactoring but kept the feature the same. 

I changed the name of the Utils class as I'm a bit annoyed by the FixedLayoutFactory. I think this factory should be kept for testings purposes only. When there's a node which doesn't have a fixed position, leaving it in 0,0 is not very interesting for the user. But that's not the topic of this PR. Let's discuss this later.

So-Fras (approved):

PR #669, opened 2025-01-20
Represent subnetworks in network-area diagrams

### LINKED ISSUE #611: Represent subnetworks in network-area diagrams

A network can be composed of several subnetworks that have been merged together. Those subnetworks are linked through tie lines.
When drawing a network-area diagram of a network composed of 2 subnetwroks, there is no way to tell which subnetwork a voltage level belongs to.

The subnetworks should be easily distinguishable from one another.

Several ideas could be tested jointly or separately:

- Represent tie lines differently from lines (by adding an element at the center of the tie lines for example, see wath is already done in SvgWriter > drawEdgeCenter)
- Draw each subnetwork separately (when using a force layout) in order to have 2 "clouds of points" on each side of the image, with the tie lines in between.
- Elements of each subnetwork could be highlighted with a different color:
image

### Describe the motivation

This feature would allow users to be able to make subnetworks out at a glance

### Extra Information

_No response_

quinarygio:

Fixes #611

- [ ] Yes
- [x] No

**Other information**:


So-Fras:
Thanks for the work!

As discussed together:

- It was not mentioned in the issue but it would be great if users could choose whether or not they want to highlight subnetworks. Maybe, via a new `SvgParameters` boolean attribute? The SvgWriter has already access to the SvgParameters. There can be a `if` condition using this new SvgParameters boolean that would trigger the duplication of lines and nodes (or not). Furthermore, this way, all the duplicated parts would be grouped together in the svg. 

- The `Graph` should not be the object to hold information on style (it's only nodes and edges) --> the `NetworkGraphBuilder` class should not be affected by this PR

- The `StyleProvider` "family" should be the one affected by the PR. In the `if` condition in the `SvgWriter` mentioned above, we would call a new specific method of the `StyleProvider` for highlight. Thus, the StyleProvider does not need to "know" whether the user wants to highlight something or not.

So-Fras (changes requested):
Thanks for the changes! I think the PR is in line with the code's "philosophy".
There is just a small typo in a name.
As for the unit test barrier, I think you should add tests to cover a few more lines and reach the 90%.

So-Fras (approved):

network-area-diagram/src/main/java/com/powsybl/nad/svg/iidm/AbstractVoltageStyleProvider.java
```diff
@@ -117,4 +125,73 @@ protected Optional<String> getBaseVoltageStyle(BranchEdge edge, BranchEdge.Side
...
+        }
+        return subnetworId;
+    }
+
+    private String getSubnetworkId(String id, ThreeWtEdge.Side side) {
+        return network.getThreeWindingsTransformer(id).getLeg(ThreeSides.valueOf(side.name())).getTerminal().getVoltageLevel().getParentNetwork().getId();
+    }
+
+    private void buildSubnetworkMaps() {
+        List<VoltageLevel> voltageLevels = getVoltageLevels();
```
flo-dup, line 178: if you only build the subnetwork map, just loop over the `network.getSubnetworks()`

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/iidm/AbstractVoltageStyleProvider.java
```diff
@@ -108,4 +114,61 @@ protected Optional<String> getBaseVoltageStyle(BranchEdge edge, BranchEdge.Side
...
+
+    @Override
+    public List<String> getHighlightThreeWtEdgStyleClasses(ThreeWtEdge edge) {
+        String subnetworkId = getSubnetworkId(edge.getEquipmentId(), edge.getSide());
+        return List.of(subnetworksHighlightMap.get(subnetworkId));
+    }
+
+    private String getSubnetworkId(BranchEdge edge, Side side) {
+        String subnetworkId = null;
+        switch (edge.getType()) {
```
flo-dup, line 138: This could be compacted, iidmSide can be reused everywhere, and terminal intermediate variable used.
```java
final TwoSides iidmSide = IidmUtils.getIidmSideFromBranchEdgeSide(side);
Terminal terminal = switch (edge.getType()) {
    case BranchEdge.LINE_EDGE -> (...)
    (...)
};
return terminal.getVoltageLevel().getParentNetwork().getId();
```

network-area-diagram/src/main/java/com/powsybl/nad/svg/iidm/AbstractVoltageStyleProvider.java
```diff
@@ -108,4 +114,61 @@ protected Optional<String> getBaseVoltageStyle(BranchEdge edge, BranchEdge.Side
...
+        return List.of(subnetworksHighlightMap.get(subnetworkId));
+    }
+
+    @Override
+    public List<String> getHighlightThreeWtEdgStyleClasses(ThreeWtEdge edge) {
+        String subnetworkId = getSubnetworkId(edge.getEquipmentId(), edge.getSide());
+        return List.of(subnetworksHighlightMap.get(subnetworkId));
+    }
+
+    private String getSubnetworkId(BranchEdge edge, Side side) {
```
flo-dup, line 136: you should take care of the case of returning null, the returned list should be empty then. Then it's easier that `getSubnetworkId` returns `Optional<String>` to handle this.

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgParameters.java
```diff
@@ -55,7 +55,7 @@ public class SvgParameters {
     private EdgeInfoEnum edgeInfoDisplayed = EdgeInfoEnum.ACTIVE_POWER;
     private double pstArrowHeadSize = 8;
     private String undefinedValueSymbol = "";
-
+    private boolean highlightSubnetworks;
```
flo-dup, line 58: while trying to complete the documentation I realized that this parameter is not properly named. It could be used to highlight anything else, it's only the default StyleProvider which highlights subnetworks. Could you rename it highlightGraph? or shadowGraph maybe? or do you have a better name in mind?

quinarygio: highlightFeature ?

network-area-diagram/src/main/java/com/powsybl/nad/svg/iidm/AbstractVoltageStyleProvider.java
```diff
@@ -123,8 +123,8 @@ public List<String> getHighlightNodeStyleClasses(Node node) {
 
     @Override
     public List<String> getHighlightSideEdgeStyleClasses(BranchEdge edge, BranchEdge.Side side) {
-        String subnetworkId = getSubnetworkId(edge, side);
-        return List.of(subnetworksHighlightMap.get(subnetworkId));
+        Optional<String> subnetworkId = getSubnetworkId(edge, side);
+        return subnetworkId.isPresent() ? List.of(subnetworksHighlightMap.get(subnetworkId.get())) : Collections.emptyList();
```
flo-dup, line 127:
```suggestion
        return getSubnetworkId(edge, side).map(id -> List.of(subnetworksHighlightMap.get(id))).orElse(Collections.emptyList());
```

network-area-diagram/src/main/java/com/powsybl/nad/svg/iidm/AbstractVoltageStyleProvider.java
```diff
@@ -133,30 +133,17 @@ public List<String> getHighlightThreeWtEdgStyleClasses(ThreeWtEdge edge) {
...
+        final TwoSides iidmSide = IidmUtils.getIidmSideFromBranchEdgeSide(side);
+        Terminal terminal = switch (edge.getType()) {
+            case BranchEdge.LINE_EDGE -> network.getLine(edge.getEquipmentId()).getTerminal(iidmSide);
+            case BranchEdge.TWO_WT_EDGE, BranchEdge.PST_EDGE -> network.getTwoWindingsTransformer(edge.getEquipmentId()).getTerminal(iidmSide);
+            case BranchEdge.DANGLING_LINE_EDGE -> network.getDanglingLine(edge.getEquipmentId()).getTerminal();
+            case BranchEdge.TIE_LINE_EDGE -> network.getTieLine(edge.getEquipmentId()).getTerminal(iidmSide);
+            case BranchEdge.HVDC_LINE_EDGE -> network.getHvdcLine(edge.getEquipmentId()).getConverterStation(iidmSide).getTerminal();
+            default -> null;
+        };
+        return terminal == null ? Optional.empty() : Optional.of(terminal.getVoltageLevel().getParentNetwork().getId());
```
flo-dup, line 146:
```suggestion
        return Optional.ofNullable(terminal).map(t -> t.getVoltageLevel().getParentNetwork().getId());
```

flo-dup (changes requested):

network-area-diagram/src/test/java/com/powsybl/nad/svg/SubnetworkHighlightTest.java
```diff
@@ -0,0 +1,488 @@
...
+    @Override
+    protected LabelProvider getLabelProvider(Network network) {
+        return new DefaultLabelProvider(network, getSvgParameters()) {
+        };
+    }
+
+    @Test
+    void testSubnetworkHighlight() {
+        Network network = createWithTieLines();
+        assertSvgEquals("/subnetwork_highlight.svg", network);
```
flo-dup, line 51: Could you add another test with a partial graph, for instance `VoltageLevelFilter.createVoltageLevelDepthFilter(network, "VLHV2", 1)`?
So that you test the case of a non-visible edge, and so the coverage will be improved.

flo-dup (changes requested):
Could you correct the signoff of the second to last commit 71816fe9 and then push-force? So that the DCO passes.

flo-dup (approved):

So-Fras (approved):

Issue #693, opened 2025-04-22
Rewrite of the layout calculation algorithm for the Network Area Diagram


NathanDissoubray:
### Describe the current behavior

The current force algorithm used to layout networks in the NAD takes a long time to run, and the result is not visually pleasing. It takes about 6 minutes on a i7-11850H and is not very readable.

![Image](https://github.com/user-attachments/assets/2e10ef00-0653-46d5-95a8-c7a2b8b490d1)

Data for the graph taken from [RTE public dataset for French Network in 2021](https://huggingface.co/datasets/rte-france/RTE7000/blob/main/2021/06/30/recollement-auto-20210630-0000-enrichi.xiidm.bz2)

### Describe the expected behavior

Using Gephi's Force Atlas 2 force diagram, we get a more pleasing result in between 30 to 45s.

![Image](https://github.com/user-attachments/assets/4ad2b380-730f-4457-a3ff-6680a76f22d3)

This needs to be nuanced a bit as Gephi states that "Gephi uses an OpenGL 3D engine to speed up graph visualization", it is unclear if that is just for the visualization or if OpenGL is also used to speed up the calculations of the force diagram. I have not taken a look at Gephi's source code myself to check, because looking at the source code of Gephi licensed under GPL 3 to then implement a similar code in powsybl, under MPL 2, would not work well for the licenses.

However, it shows that a more visually pleasing result in less time is something feasible (Gephi is also written in Java), even with the fact that some more details need to be shown on the lines of the NAD than what Gephi can do, the general layout is still clearer than the current force layout diagram of the NAD.

### Describe the motivation

The NAD allows the display of a small network but becomes rapidly unreadable for bigger networks, which makes it unsuitable for this kind of display currently, when in theory it could be a great tool for doing so.

## Papers used and ideas

An initial implementation of the Atlas 2 Force algorithm would be based on the paper: [ForceAtlas2, a Continuous Graph Layout Algorithm for Handy Network Visualization Designed for the Gephi Software](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0098679), Jacomy M, Venturini T, Heymann S, Bastian M (2014)

[...]

PR #695, opened 2025-05-14
Refactor Force Layout

NathanDissoubray:

Refactoring using the structure described in #693 
This does not implemented Atlas2 yet, it makes the structure for it and rewrites the old force layout to follow that structure

This PR is a refactor of the Springy layout used in the NAD to place nodes and edges.

The behavior is the same before and after the refactor (using the tests in the NAD to check that the output is the same)

- [ ] Yes
- [x] No

We could decide to make some methods deprecated though, like the getSprings of forceLayout, that would allow the removal of the Spring class.

**Other information**:

The test coverage is about 50% in the module of the new class, however most of the class are tested by the tests in the NAD. Those are global tests though, not unit tests, so I'm thinking of adding more unit tests for util, so I will mark this as draft for now.


diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/forceparameter/ForceParameter.java
flo-dup, line 1: As discussed, this is an interface without any methods (called a "marker interface") and it's usually a good hint that there's a design problem

NathanDissoubray:
speedup_comparison
A comparison of the speed before and after the refactor. Two things that are worth noting:
- the more points in the graph, the more the new version seem to do better than the old version, which is good. It seems to stabilize at 70%~75% speedup for 8k points graph, but I did not test for more points
- the scale for the time is logarithmic, and both curves seem linear, which would imply that the complexity of springy is exponential, rather than the theoretical quadratic complexity, which is bad.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        for (E edge : graph.edgeSet()) {
+            V source = graph.getEdgeSource(edge);
+            V target = graph.getEdgeTarget(edge);
+            if (source != target) {
+                locSimpleGraph.addEdge(graph.getEdgeSource(edge), graph.getEdgeTarget(edge));
+            }
+        }
+        this.simpleGraph = locSimpleGraph;
+        this.graph = graph;
```
flo-dup: Do you need to keep it? I think we can keep only the simple one

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/forceparameter/IntensityEffectFromFixedNodesParameters.java
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+package com.powsybl.diagram.util.forcelayout.forces.forceparameter;
```
flo-dup:
```suggestion
package com.powsybl.diagram.util.forcelayout.forces.parameters;
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/layoutsparameters/AbstractLayoutParameters.java
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+package com.powsybl.diagram.util.forcelayout.layouts.layoutsparameters;
```
flo-dup:
```suggestion
package com.powsybl.diagram.util.forcelayout.layouts.parameters;
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
```diff
@@ -0,0 +1,20 @@
...
+package com.powsybl.diagram.util.forcelayout.forces;
+
+import com.powsybl.diagram.util.forcelayout.geometry.ForceGraph;
+import com.powsybl.diagram.util.forcelayout.geometry.Point;
+import com.powsybl.diagram.util.forcelayout.geometry.Vector2D;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public abstract class AbstractForce<V, E> {
```
flo-dup:
```suggestion
public interface Force<V, E> {
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
```diff
@@ -0,0 +1,20 @@
...
+    public abstract Vector2D calculateForce(V forThisVertex, Point correspondingPoint, ForceGraph<V, E> forceGraph);
```
flo-dup:
```suggestion
    public abstract Vector2D apply(V forThisVertex, Point correspondingPoint, ForceGraph<V, E> forceGraph);
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
```diff
@@ -0,0 +1,197 @@
...
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class ForceGraph<V, E> {
```
flo-dup: We could rename this, there's nothing about forces here, is it?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
```diff
@@ -0,0 +1,197 @@
...
+
+        printWriter.close();
+    }
+
+    private Point getPointWithVertex(V vertex) throws NoSuchElementException {
+        Point point = movingPoints.get(vertex);
+        if (point != null) {
+            return point;
+        } else {
+            point = fixedPoints.get(vertex);
```
flo-dup: Exception is rather used for unexpected results, null is fair as an empty result.
```suggestion
            return fixedPoints.get(vertex);
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/LayoutAlgorithmRunner.java
```diff
@@ -0,0 +1,97 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Function;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class LayoutAlgorithmRunner<V, E> {
+    private AbstractSetup<V, E> setup;
+    private AbstractLayoutAlgorithm<V, E> layoutAlgorithm;
```
flo-dup: We rather use interfaces than abstract classes when possible
- AbstractSetup -> Setup interface
- AbstractLayoutAlgorithm -> LayoutAlgorithm interface
- ...

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/GravityForceByEdgeNumber.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public GravityForceByEdgeNumber(IntensityParameter forceParameter) {
+        this.forceParameter = forceParameter;
+    }
+
+    @Override
+    public Vector2D calculateForce(V forThisVertex, Point correspondingPoint, ForceGraph<V, E> forceGraph) {
+        // magnitude = k * (deg (point) + 1)
+        // with deg(p) the degree of p, ie the number of connected nodes, that is to say the number of edges
+        // this means less connected points will end more on the sides of the graph
```
flo-dup: hmm I don't really see the link, could you elaborate a bit more?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/LayoutAlgorithmRunner.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Function;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class LayoutAlgorithmRunner<V, E> {
```
flo-dup: We should try to make a better difference in classes names between (Abstract)LayoutAlgorithm and LayoutAlgorithmRunner, which does more than just run the LayoutAlgorithm

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
NathanDissoubray: That was originally to still represent multiple edges when exporting to SVG, but since we are not curving the lines when there are multiple edges between two vertex, we don't see anything different; so it's ok to delete it I think

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
NathanDissoubray: I wouldn't call it "apply" because to me "calculate" returns something, whereas I expected a function called "apply" to work by edge effect and to not return anything

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
NathanDissoubray: No there is not, `LayoutData` ?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
NathanDissoubray: Will use an optional I think

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/GravityForceByEdgeNumber.java
NathanDissoubray: The gravity force is multiplied by the degree of the vertex. That means that a node with 9 connections will have a gravity force 10 times stronger than a node with no connections (which will have a gravity of 1)
Since there are repulsion forces at work too, nodes that are less connected get pushed towards the side of the graph by nodes that are more connected and are thus more attracted to the center

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/LayoutAlgorithmRunner.java
NathanDissoubray: I'm not too sure about the name here
Maybe `ForceLayoutRunner` would be clearer ? Since the package is forcelayout, it would also make it clearer that this is the main entrypoint

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/forceparameter/IntensityEffectFromFixedNodesParameters.java
NathanDissoubray: Done in 643c66188ac0

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/layoutsparameters/AbstractLayoutParameters.java
NathanDissoubray: Done in d46ff3507879

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
NathanDissoubray: Done in 577727f34350

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
NathanDissoubray: Done in ada593bf629d

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
NathanDissoubray: Done in 19fd62b6275e

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
NathanDissoubray: Also 797cfdb4ffe1 because when I changed from abstract to interface, I forgot to actually rename the class (interestingly, there is no rule in CheckStyle that prevents an interface from being called AbstractSomething)

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/LayoutAlgorithmRunner.java
NathanDissoubray: Done in 2e9174ff3f80 bbe6d6f17236 and 02d899fd8ca7

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/LayoutAlgorithmRunner.java
NathanDissoubray: Ah but that would make it confusing with ForceLayout (which is just Springy actually)

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
flo-dup: Well, at least in java, apply is often used for methods returning a result, see java.util.Function::apply method:
```java
    /**
     * Applies this function to the given argument.
     *
     * @param t the function argument
     * @return the function result
     */
    R apply(T t);
```

NathanDissoubray: Mmh I see, guess I'm just not used to this notation, will change it then

flo-dup: Could you also rename forThisVertex into vertex? and correspondingPoint into point? For conciseness. Both "for this" and "corresponding" are implicit in almost all parameters.

NathanDissoubray: Done in 2c8feacbac63 (to rename calculateForce into apply)

NathanDissoubray: > Could you also rename forThisVertex into vertex? and correspondingPoint into point? For conciseness. Both "for this" and "corresponding" are implicit in almost all parameters.

vertex instead of forThisVertex seems ok to me, but correspondingPoint states that this is the point in link with the vertex. If we just say "point", I feel like it says that it can be any point, not just the 2D point corresponding to the vertex in the graph
Or I rename to point but I add javadoc above to explain that the parameter point is the one corresponding to the given vertex, but I feel like having the name be clear directly would be better

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/ForceGraph.java
flo-dup: I think LayoutContext is more appropriate, `Data` is usually a keyword which is avoided, it's too generic and it's also suggesting raw immutable input or output.

NathanDissoubray: Done in 54031550d85f

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/AbstractForce.java
flo-dup: you'd rather add it in the javadoc only and keep it concise in the variable name. Naming a variable `point` alone means implicitly **The** point. If that's ambiguous, then naming it correspondingPoint does not make it clearer as we don't know to which object it corresponds.

NathanDissoubray: That makes sense, done in 7025d542a6ed

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Canvas.java
```diff
@@ -26,10 +30,10 @@ public Canvas(BoundingBox boundingBox, double height, double margin) {
         this.margin = margin;
     }
 
-    public Vector toScreen(Vector position) {
+    public Vector2D toScreen(Vector2D position) {
         double screenX = (position.getX() - boundingBox.getLeft()) * scale + margin;
-        double screenY = (position.getY() - boundingBox.getTop()) * scale + margin;
-        return new Vector(screenX, screenY);
+        double screenY = (position.getY() - boundingBox.getBottom()) * scale + margin;
```
flo-dup: you're considering the 0,0 position at the bottom left then? that 0,0 position is at the top-left corner in most computer graphics frameworks

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/SpringyLayout.java
```diff
@@ -0,0 +1,130 @@
...
+import com.powsybl.diagram.util.forcelayout.layouts.parameters.SpringyParameters;
+import org.jgrapht.graph.DefaultEdge;
+import org.jgrapht.graph.SimpleGraph;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
```
flo-dup: I think we should mention the springy.js inspiration for this layout in this file as now the code is mainly here. And as we're now quite far from its js code I think we could keep our copyright header.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
flo-dup, line 1: Therefore we could change the copyright here, and speak of "inspired" instead of "taken". This justifies renaming this class (breaking change!)

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.powsybl.diagram.util.forcelayout.geometry.LayoutContext;
+
+import java.util.Random;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public interface Setup<V, E> {
+    void setup(LayoutContext<V, E> layoutContext, Random random);
```
flo-dup: Asking for a random object is too specific to the implementation you have in mind. I can think of many setup which don't need randomness.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
```diff
@@ -0,0 +1,19 @@
...
+package com.powsybl.diagram.util.forcelayout.setup;
+
+import com.powsybl.diagram.util.forcelayout.geometry.LayoutContext;
+
+import java.util.Random;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public interface Setup<V, E> {
```
flo-dup: `setup.setup()` is redundant, you'd better rename the method.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Canvas.java
NathanDissoubray: Actually I was considering (0, 0) to be the screen center in the 2D space (I have points with negative coordinates), but it's different for the canvas indeed, we can do a transformation to consider top left instead of bottom left, but that means axis are inverted on the y compared to the 2D space:
       2D space
```
       ^ y
       |
--- (0, 0) ---> x
       |
       |
```  

Canvas space
```
(0, 0) ---> x
|
|
v y
```

Something like that ?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
NathanDissoubray: What would we rename ForceLayout to ? Springy would make it confusing because we already have a springylayout

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
NathanDissoubray: setup.do() ?
setup.run() ? I think that one is more in line with LayoutAlgorithmRunner.run() but that could also make it confusing

Most setups that inherit from this interface are not just named setup, so in the code that uses that it's more like SimpleSetup.setup()

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
NathanDissoubray: I see what you mean, however I checked the extend / implement of Random and it's object / serializable
If we want to make a setup that does not require a random, we would need something more general, but given the supertype that seems difficult.

There are indeed, setup that do not require a random, but are we really gonna use those in the force layout ? For example, fixed position does not require a random, but we are not gonna do a force layout if we already know we are doing a force position. Which setups are we actually gonna implement in force layout, that do something different (and have a notable effect on the graph compared to a random setup) that do not require a random ?

NathanDissoubray: I just realized we could give the random as a field of the object when building the instance of the class that implements interface, but I still wonder if we are gonna make setups that do not require a random.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/SpringyLayout.java
NathanDissoubray: Done in 42166cbc78bd

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
NathanDissoubray: Done in ddf312053c1b

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
NathanDissoubray: Done in e28659866423

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Canvas.java
NathanDissoubray: Done in 3d8ea584882f

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
```diff
@@ -0,0 +1,135 @@
...
+    public Layout(Setup<V, E> setup, LayoutAlgorithm<V, E> layoutAlgorithm) {
+        this.setup = setup;
+        this.layoutAlgorithm = layoutAlgorithm;
+    }
+
+    /**
+     * Get the default Springy algorithm, with {@link SpringySetup} and {@link SpringyLayout}
+     * @return a ready to run springy algorithm, with default parameters
+     */
+    public static <V, E> Layout<V, E> getSpringyDefaultLayout() {
```
flo-dup: As already mentioned we can avoid naming classes and methods springy now that the code got even further away from it. `getBasicForceLayout` for instance?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/CoulombForce.java
```diff
@@ -0,0 +1,61 @@
...
+                coulombBetweenPoints(resultingForce, point, otherFixedPoint);
+            }
+        }
+        return resultingForce;
+    }
+
+    private void coulombBetweenPoints(Vector2D resultingForce, Point correspondingPoint, Point otherPoint) {
+        Vector2D force = Vector2D.calculateVectorBetweenPoints(otherPoint, correspondingPoint);
+        double magnitude = force.magnitude();
+        /// this is the contracted version of calculating Coulomb, if we take V as the vector from otherPoint to correspondingPoint, then
```
flo-dup: The inline comments should start with two slash and not three

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/SpringForce.java
```diff
@@ -0,0 +1,83 @@
...
+            Point pointSource = Objects.requireNonNullElseGet(layoutContext.getMovingPoints().get(edgeSource), () -> layoutContext.getInitialPoints().get(edgeSource));
+            Point pointTarget = Objects.requireNonNullElseGet(layoutContext.getMovingPoints().get(edgeTarget), () -> layoutContext.getInitialPoints().get(edgeTarget));
+            if (pointSource != pointTarget) { // no use in force layout to add loops
+                springs.put(edge, new SpringParameter(DEFAULT_STIFFNESS, simpleGraph.getEdgeWeight(edge)));
+            }
+        }
+        forceParameter.setSprings(springs);
+    }
+
+    /// This is Hooke's Law
```
flo-dup: Either convert it to javadoc comment styling or put it inside the method as inline comment

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/Vector2D.java
```diff
@@ -0,0 +1,89 @@
...
+        this.divideBy(this.magnitude());
+    }
+
+    public static Vector2D calculateVectorBetweenPoints(Point from, Point towards) {
+        Vector2D direction = new Vector2D(towards.getPosition());
+        direction.subtract(from.getPosition());
+        return direction;
+    }
+
+    /// Calculate the unit vector that goes from This point, pointing in the direction of towards
```
flo-dup: Same comment

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SpringySetup.java
```diff
@@ -0,0 +1,45 @@
...
+import com.powsybl.diagram.util.forcelayout.geometry.Vector2D;
+
+import java.util.Random;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class SpringySetup<V, E> extends SimpleSetup<V, E> {
+    public SpringySetup(Random random) {
+        this.random = random;
```
flo-dup: You should use `super(setup)` instead, which allows protecting random field by making it private and final

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SpringySetup.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public SpringySetup() {
+    }
```
flo-dup: Use `super(new Random(DEFAULT_SEED))` instead

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Objects;
+import java.util.Random;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class SimpleSetup<V, E> implements Setup<V, E> {
+    @SuppressWarnings("java:S2245")
+    protected Random random = new Random(3L);
+    protected Vector2D center;
```
flo-dup: `center` does not need to be a field. It could be a local value inside `SpringySetup` instead.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
```diff
@@ -0,0 +1,51 @@
...
+    protected double scale = 1;
```
flo-dup: To make it private and final, you can add a constructor with scale as parameter.

diagram-util/src/test/java/com/powsybl/diagram/util/forcelayout/Helpers.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class Helpers {
+    public String toString(String resourceName) {
```
flo-dup: Could be a static method. Helpers should be renamed to have a more explicit name (something like ResourceUtils?)

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
```diff
@@ -0,0 +1,135 @@
...
+ * Alternatively, use {@link #getStablePosition(Object)} to get the position of the point corresponding to the vertex object, this allows to write your own output<br>
+ * Note that you do not need to create a new <code>Layout</code> to run on a different <code>layoutContext</code>, you can just re-use the same layout object and start with run again
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class Layout<V, E> {
+    private final Setup<V, E> setup;
+    private final LayoutAlgorithm<V, E> layoutAlgorithm;
+    private boolean hasBeenExecuted = false;
+    private LayoutContext<V, E> layoutContext;
+    private Vector2D center = new Vector2D(0, 0);
```
flo-dup: The layout should not hold these two parameters.
- center would make more sense as a parameter of the LayoutContext,
- layoutContext is given by the caller, so there's no need to keep it inside Layout class: the caller has access to it; hence, toSvg and getStablePosition could be directly taken from the context

Then, hasBeenExecuted parameter could be moved also to LayoutContext, if we need to keep it - do we?

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(Layout.class);
+
+    /**
+     * Create a new Layout instance with the provided setup and algorithm
+     * @param setup the setup to be used
+     * @param layoutAlgorithm the algorithm to place the points once the setup has been applied
+     */
+    public Layout(Setup<V, E> setup, LayoutAlgorithm<V, E> layoutAlgorithm) {
+        this.setup = setup;
+        this.layoutAlgorithm = layoutAlgorithm;
```
flo-dup: In a public method which takes objects as parameters, you should always check that they are non-null first - except if you need to handle the null-case everywhere you're manipulating it
```suggestion
        this.setup = Objects.requireNonNull(setup);
        this.layoutAlgorithm = Objects.requireNonNull(layoutAlgorithm);
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/SpringForce.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Objects;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class SpringForce<V, E> implements Force<V, E> {
+    private final SpringContainer<DefaultEdge> forceParameter;
+    private static final double DEFAULT_STIFFNESS = 100.0;
+
+    public SpringForce(SpringContainer<DefaultEdge> forceParameter) {
```
flo-dup: this is only called with an empty container, so replace this constructor with a no-args one

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/parameters/SpringContainer.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.powsybl.diagram.util.forcelayout.forces.parameters;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class SpringContainer<E> {
```
flo-dup: Why not wrapping a Map, but then do not give access to that map, or the class does not bring much.
If you want to keep it, introduce a `put(edge, parameter)` and `getParameter(edge)` method inside.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
NathanDissoubray: Should I remove the copyright mention in ForceLayout altogether as well ?

NathanDissoubray: Nevermind forget that, the long copyright mention is in ForceLayout which we deprecated

NathanDissoubray: 757c674b6855

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/CoulombForce.java
NathanDissoubray: 6b8d61f5aadf

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/SpringForce.java
NathanDissoubray: 6b8d61f5aadf

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/Vector2D.java
NathanDissoubray: 6b8d61f5aadf

diagram-util/src/test/java/com/powsybl/diagram/util/forcelayout/Helpers.java
NathanDissoubray: 8a5125683166

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
NathanDissoubray: 1f043043e154

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/SpringForce.java
NathanDissoubray: 630c2dc352b1

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/parameters/SpringContainer.java
NathanDissoubray: 35b57fda39a2
I removed the class instead, it was not needed

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SpringySetup.java
NathanDissoubray: e208209d7836

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SpringySetup.java
NathanDissoubray: e208209d7836
you will need to resolve the problem with the insecure pseudo-random in Sonar because I do not have the permissions to do it

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
NathanDissoubray: I can't do that because in SimpleBarycenterSetup, I need the layoutContext to calculate the scale, and that comes only after creating the Setup object

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
NathanDissoubray: 85510c5c3d7d
removed hasBeenExecuted too, it was not needed

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/forces/GravityForceByEdgeNumber.java
flo-dup: quite obvious, sorry, got completely lost with "gravity" naming!

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/parameters/BasicParameters.java
flo-dup, line 1: You should rather name it BasicLayoutParameters, here it sounds like the parameters are basic :slightly_smiling_face:

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/LayoutContext.java
```diff
@@ -51,47 +54,75 @@ public ForceGraph(Graph<V, E> graph) {
...
 
+    /**
+     * @return a Map vertex/point where the points had an initial position given to them before the setup
+     */
     public Map<V, Point> getInitialPoints() {
         return initialPoints;
     }
 
+    /**
+     * @return a set of all the vertex where the point corresponding to those is fixed. Note that this does not give the corresponding points directly
```
flo-dup: Replace everywhere "the vertex" by "the vertices" when there are several

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/BasicLayout.java
```diff
@@ -0,0 +1,105 @@
...
+            this.forces.add(new AttractToCenterForceLinear<>(layoutParameters.getRepulsion() / 200));
+        }
+        this.layoutParameters = layoutParameters;
+    }
+
+    @Override
+    public void run(LayoutContext<V, E> layoutContext) {
+        Objects.requireNonNull(layoutContext);
+        //Initialize Spring force
+        this.forces.get(0).init(layoutContext);
```
flo-dup: All forces have init, even if there's nothing done in it, it makes more sense to call it for everyone

```suggestion
        this.forces.forEach(f -> f.init(layoutContext));
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Set up the layout context so the layout algorithm can run on it properly<br>
+     * An implementation of a setup is expected to do at least the following:<br>
+     * - fill the <code>fixedPoints</code> and <code>movingPoints</code> Map of <code>layoutContext</code> depending on if the vertex are in <code>fixedNodes</code> or not<br>
+     * - use <code>initialPoints</code> to give a starting position to all the points, or create a position if not initial position is given<br>
```
flo-dup:
```suggestion
     * - use <code>initialPoints</code> to give a starting position to all the points, or create a position if no initial position is given<br>
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
```diff
@@ -0,0 +1,52 @@
...
+import com.powsybl.diagram.util.forcelayout.geometry.Point;
+import com.powsybl.diagram.util.forcelayout.geometry.Vector2D;
+
+import java.util.Objects;
+import java.util.Random;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class SimpleSetup<V, E> implements Setup<V, E> {
```
flo-dup: Choose between Simple and Basic to have the same keyword for basic/simple classes.

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/BasicLayout.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * The following algorithm is a force layout algorithm.
+ * It seeks to place the nodes of a graph in such a way that the nodes are well spaced and that there are no unnecessary crossings.
+ * The algorithm uses an analogy with physics where the nodes of the graph are particles with mass and the edges are springs.
+ * Force calculations are used to place the nodes.
+ * The algorithm is inspired from: https://github.com/dhotson/springy
+ *
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class BasicLayout<V, E> implements LayoutAlgorithm<V, E> {
```
flo-dup: It's a layout based on forces, so we should name it BasicForceLayout. Which makes it confusing with nad similar named class :slightly_frowning_face:

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
```diff
@@ -0,0 +1,72 @@
...
+    public Layout(Setup<V, E> setup, LayoutAlgorithm<V, E> layoutAlgorithm) {
+        this.setup = Objects.requireNonNull(setup);
+        this.layoutAlgorithm = Objects.requireNonNull(layoutAlgorithm);
+    }
+
+    /**
+     * Get the default Basic algorithm, with {@link SimpleBarycenterSetup} and {@link BasicLayout}
+     * @return a ready to run basic algorithm, with default parameters
+     */
+    public static <V, E> Layout<V, E> getBasicDefaultLayout() {
```
flo-dup: Rather create as "get" sounds like getting a singleton. And I woudln't name it default as the default should rather be another implementation.
```suggestion
    public static <V, E> Layout<V, E> createBasicForceLayout() {
```

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
flo-dup: SquareRandomSetup > BarycenterSquareRandomSetup

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
```diff
@@ -46,265 +49,108 @@
  * The algorithm uses an analogy with physics where the nodes of the graph are particles with mass and the edges are springs.
  * Force calculations are used to place the nodes.
  *
- * The algorithm is taken from: https://github.com/dhotson/springy
+ * The algorithm is inspired from: https://github.com/dhotson/springy
+ * @deprecated Use {@link com.powsybl.diagram.util.forcelayout.Layout} instead <br>
+ * The equivalent of: <br>
+ * {@code new ForceLayout(graph).execute();}<br>
+ * would be <br>
+ * {@code new Layout(new SimpleBarycenterSetup<>(), new BasicLayout<>(new BasicParameters.Builder().build()).run(new LayoutContext<>(graph));}
```
flo-dup: Refer to the static method you created on purpose instead!

NathanDissoubray: I forgot to update the doc indeed

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/parameters/BasicParameters.java
NathanDissoubray: 1fe1283feb0a

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/geometry/LayoutContext.java
NathanDissoubray: e3782771bb7e

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/BasicLayout.java
NathanDissoubray: Added it as a default method of the interface to be used by other class implementing the interface too
448e3f3d7ae8

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/Setup.java
NathanDissoubray: 8523f7b49a9d

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/setup/SimpleSetup.java
NathanDissoubray: 8fb89d0fe3e7

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/Layout.java
NathanDissoubray: 141e944f1345

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/ForceLayout.java
NathanDissoubray: f525514a6e67

diagram-util/src/main/java/com/powsybl/diagram/util/forcelayout/layouts/BasicLayout.java
NathanDissoubray: Should we open an issue to rename BasicForceLayout in NAD ? Or keep it like that ?

flo-dup: I think we'll have a better idea how to change it in NAD once you add another implementation. But yes you could create the issue even if we don't know clearly what to do instead.

flo-dup (approved):

PR #697, opened 2025-06-11
[NAD] Display injections on diagram if requested

### LINKED ISSUE #613: Display power production and power consumption at each voltage level node

In network-area diagrams, it is not currently possible to see which voltage level produces or consumes power.

The active power produced and consumed at each voltage level node should be displayed:
- in a first attempt, as text in the voltage level text box for example
- afterwards along 2 small icons (one for production, one for consumption):
image

Displaying those pieces of information should be an option.

### Describe the motivation

It would help users see at once which node is more of a consumer or a producer.

### Extra Information

_No response_

flo-dup:

Yes, fixes #613

Feature

The power consumption/production can be displayed only for the whole voltage level in the text box

All the injections are displayed inside the SVG based on the provided component library, similarly to what's done in powsybl-single-line-diagram.
Note that grounds (might be easily added), HVDC (seen as a branch), dangling lines (seen as a voltage level node) and busbar sections (no usecase?) are not displayed.

- [x] Yes
- [ ] No

**Other information**:
Capture d’écran du 2025-06-11 17-22-03

### DIFF
--- a/network-area-diagram/src/test/java/com/powsybl/nad/svg/OverridingStyleTest.java
+++ b/network-area-diagram/src/test/java/com/powsybl/nad/svg/OverridingStyleTest.java
@@ -81,7 +81,6 @@ class OverridingStyleTest extends AbstractTest {
     void testEurope() {
         Network network = Network.read("simple-eu.uct", getClass().getResourceAsStream("/simple-eu.uct"));
         assertSvgEquals("/simple-eu_overridden_styles.svg", network);
-        assertSvgEquals("/simple-eu_overridden_styles.svg", network);
     }
 
     @Test


network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgStAXResult.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javanet.staxutils.ContentHandlerToXMLStreamWriter;
+
+import javax.xml.stream.XMLStreamWriter;
+import javax.xml.transform.sax.SAXResult;
+
+/**
+ * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
+ */
+public class SvgStAXResult extends SAXResult {
```
So-Fras: Is it possible to add a comment in the class to explain what it is about? Even though we talked about it, it is still unclear to me.

So-Fras: And why is the class named `SvgStAXResult` instead of `SvgSAXResult`?

flo-dup: It was named SvgStAX because I first extended StAXRresult, but SAXResult was eventually enough. Renaming it!

flo-dup: In fact I simplified it a bit, the SAXResult itself could be used with the specific ContentHandler class.

So-Fras (approved):
Thanks for the changes!

So-Fras (dismissed):

PR #712, opened 2025-07-15
[SLD] Fix missing style class for bus legend

flo-dup:

No

Bug fix

With TopologicalStyleProvider, circle in bus legend info is black instead of the bus color 

Circle in bus legend info takes the colour of the corresponding bus

- [ ] Yes
- [x] No

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/AbstractTestCase.java
+++ b/single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/AbstractTestCase.java
@@ -122,7 +122,7 @@ public abstract class AbstractTestCase {
             SingleLineDiagram.draw(graph, writer, NullWriter.nullWriter(), componentLibrary, layoutParameters, svgParameters, labelProvider, styleProvider);
 
             if (debugSvgFiles) {
-                writeToFileInDebugDir(filename, writer);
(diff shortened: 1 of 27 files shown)


single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/AbstractTestCase.java
```diff
@@ -122,7 +122,7 @@ public String toSVG(Graph graph, String filename, SldComponentLibrary componentL
             SingleLineDiagram.draw(graph, writer, NullWriter.nullWriter(), componentLibrary, layoutParameters, svgParameters, labelProvider, styleProvider);
 
             if (debugSvgFiles) {
-                writeToFileInDebugDir(filename, writer);
+                writeToFileInDebugDir(filename.replace(".svg", "_new.svg"), writer);
```
So-Fras: Is it a modification you'd like to keep or was it just for debug purposes?
I am fine with the modification if you want to keep it because sometimes, it is hard to distinguish which file is which.

flo-dup: woops that was only for comparing with reference, didn't mean to commit... but yes maybe we should add some parameter which enables this or not to be able to quickly see the changes

So-Fras (dismissed):

PR #721, opened 2025-07-25
Atlas 2 force layout

NathanDissoubray:

In link with #693 
Closes #574 (duplicate)

Using the recent refactor of the force layout in diagram-util, write a new algorithm that is better both in time and visual than the basic force layout, namely Atlas2, based on the paper "ForceAtlas2, a Continuous Graph Layout Algorithm for Handy Network Visualization Designed for the Gephi Software"

We currently only have the BasicForceLayoutAlgorithm

Add Atlas2ForceLayoutAlgorithm

- [ ] Yes
- [x] No

**Other information**:

Example images will be added later to compare between BasicForceLayout and Atlas2. Some speed comparison graph will also be added.


diagram-util/src/main/java/com/powsybl/diagram/util/layout/postprocessing/NoPostProcessing.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+package com.powsybl.diagram.util.layout.postprocessing;
+
+import com.powsybl.diagram.util.layout.geometry.LayoutContext;
+
+/**
+ * Used when no post-processing is required
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class NoPostProcessing<V, E> implements PostProcessing<V, E> {
```
flo-dup:
```suggestion
public class NoOpPostProcessing<V, E> implements PostProcessing<V, E> {
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
```diff
@@ -18,15 +18,17 @@ public final class BasicForceLayoutParameters {
...
 
     private final int maxSteps;
     private final double minEnergyThreshold;
     private final double deltaTime;
     private final double repulsion;
     private final double friction;
     private final double maxSpeed;
-    private final boolean repulsionForceFromFixedPoints;
-    private final boolean attractToCenterForce;
+    private final boolean activateRepulsionForceFromFixedPoints;
```
flo-dup:
```suggestion
    private final boolean repulsionForceFromFixedPointsActivated;
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,229 @@
...
+/// the Gephi Software. PLoS ONE 9(6): e98679. doi:10.1371/journal.pone.0098679
+/// The paper requires its usage to credit the original author and the source, DO NOT REMOVE THE ABOVE REFERENCE
+/// Some parts of the code do not directly come from the paper and are instead found through experimenting, to find something that works best
+/// The parts that are not inside the original paper are:
+/// - the choice of a starting global speed for the graph
+/// - a maximum decrease ratio in global speed between each step
+/// - a "quick to calculate" stopping condition
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
```
flo-dup: The comments mentioning the ForceAtlas2 paper should be included in this documentation comment, just above the author line with your name.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/Layout.java
```diff
@@ -67,6 +90,10 @@ public void run(LayoutContext<V, E> layoutContext) {
         long setupEnd = System.nanoTime();
         LOGGER.info("Setup took {} s", (setupEnd - start) / 1e9);
         layoutAlgorithm.run(layoutContext);
-        LOGGER.info("Layout calculations took {} s", (System.nanoTime() - setupEnd) / 1e9);
+        long algorithmEnd = System.nanoTime();
+        LOGGER.info("Layout calculations took {} s", (algorithmEnd - setupEnd) / 1e9);
+        postProcessing.run(layoutContext);
```
flo-dup: Avoid timing logic duplication by introducing a runAndMeasure method, something like
```java
private void runAndMeasure(String phaseName, Runnable phase) {
    long start = System.nanoTime();
    phase.run();
    long end = System.nanoTime();
    LOGGER.info("{} took {} s", phaseName, (end - start) / 1e9);
}
```

network-area-diagram/src/main/java/com/powsybl/nad/layout/LayoutParameters.java
```diff
@@ -15,6 +15,7 @@
 public class LayoutParameters {
     private boolean textNodesForceLayout = false;
     private Point textNodeFixedShift = new Point(100, -40);
+    //TODO should this be deprecated since the maxSteps is now a parameter directly in diagram-util ?
```
flo-dup: For now I'd keep it.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
```diff
@@ -27,8 +29,8 @@ public final class BasicForceLayoutParameters {
...
```
flo-dup:
```suggestion
    private final boolean repulsionForceFromFixedPointsActivated;
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
flo-dup, line 37: The comments mentioning the ForceAtlas2 paper should be included in this documentation comment, just above the author line with your name, so that the article appears in the javadoc

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,229 @@
...
+
+        for (Point point : layoutContext.getMovingPoints().values()) {
+            previousForces.put(point, new Vector2D());
+            swingMap.put(point, 0.);
+        }
+        int i = 0;
+        int stoppingStep = layoutParameters.getMaxSteps();
+        boolean changedStoppingStep = false;
+
+        while (i < stoppingStep) {
```
flo-dup: there's now a new parameter asked by users which should be taken into account: timeoutSeconds. Problem is, the timeout should be for the full layout process. As a first step, we could only consider the algorithm part, like it's done in the basic layout. And not consider the global time in this PR, as this would probably lead to more changes.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
```diff
@@ -0,0 +1,272 @@
...
+        /**
+         * By how many iteration (in percent) you want to increase the run once the stopping condition was met.
+         * The stopping condition generally stops when the graph is "good enough", but for specific use you might want a graph that looks better.
+         * Increasing the number of iterations past the stopping condition will increase the visual quality of the graph, but it will also take longer.
+         * This coefficient will also be directly the increase in runtime (e.g. if you use 10%, you will have a 10% longer runtime).
+         * Default is {@value DEFAULT_ITERATION_NUMBER_INCREASE_PERCENT}
+         * @param iterationNumberIncreasePercent how many more iterations (in percent) you want the run to last
+         * @return the instance of this Builder with the `iterationNumberIncreasePercent` changed
+         */
+        public Builder withIterationNumberIncreasePercent(double iterationNumberIncreasePercent) {
```
flo-dup: Remove this parameter and the "changedStoppingStep" logic, the good way to go for increasing the number of steps would be to have a stopping condition interface. This looks too much like a hack.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,229 @@
...
+        // this is not part of Atlas2's paper
+        final double stoppingGlobalGraphSpeed = NORMALIZED_STOPPING_VALUE * Math.pow(graphSize, NORMALIZATION_POWER);
+
+        for (Point point : layoutContext.getMovingPoints().values()) {
+            previousForces.put(point, new Vector2D());
+            swingMap.put(point, 0.);
+        }
+        int i = 0;
+        int stoppingStep = layoutParameters.getMaxSteps();
+        boolean changedStoppingStep = false;
```
flo-dup: As mentioned, remove this. If we need to adjust the stopping step, this should not appear here in the algorithm. We should have instead a isStable specific implementation, with something like a StoppingCriteria interface. Let's remove this out of the scope of current PR.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,229 @@
...
+            for (Map.Entry<V, Point> entry : layoutContext.getMovingPoints().entrySet()) {
+                Point point = entry.getValue();
+                for (Force<V, E> force : forces) {
+                    Vector2D resultingForce = force.apply(entry.getKey(), point, layoutContext);
+                    point.applyForce(resultingForce);
+                }
+                // calculate swg(n) for each node the swing of the node
+                // at the same time calculate tra(n) the traction of the node
+                // we can also calculate swg(G) and tra(G) the swing and traction of the graph
+                int vertexDegreePlusOne = layoutContext.getSimpleGraph().degreeOf(entry.getKey()) + 1;
```
flo-dup:
```suggestion
                int weight = layoutContext.getSimpleGraph().degreeOf(entry.getKey()) + 1;
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,229 @@
...
+            }
+            if (graphSwing == 0) {
+                // that means that all the points are not moving anymore, or are diverging very fast
+                break;
+            }
+            // calculate s(G) the global speed of the graph
+            // this speed should not be less than a certain amount of the previous graph speed
+            // the graph speed should not be more than a certain amount of the previous graph speed
+            // calculate given the swing tolerance and check it's being between those bounds
+            newGraphSpeed = Math.max(
```
flo-dup: You can now use `Math.clamp` instead

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractByEdgeNumberForce.java
```diff
@@ -0,0 +1,27 @@
...
+import com.powsybl.diagram.util.layout.geometry.LayoutContext;
+import com.powsybl.diagram.util.layout.geometry.Point;
+
+import java.util.Map;
+
+/**
+ * Abstract class for all the forces that need the number of edges of a point for its calculations
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public abstract class AbstractByEdgeNumberForce<V, E> implements Force<V, E> {
```
flo-dup: You're using here ByEdgeNumber but sometimes degree instead. Keep degree keyword everywhere, it aligns with graph terminology.
```suggestion
public abstract class AbstractDegreeBasedForce<V, E> implements Force<V, E> {
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceByEdgeNumberNoOverlapLinear.java
```diff
@@ -0,0 +1,97 @@
...
+        // using default value, will change later
+        this.repulsionZoneRadius = pointSize * repulsionZoneRatio;
+    }
+
+    @Override
+    public void init(LayoutContext<V, E> layoutContext) {
+        // init point size
+        super.init(layoutContext);
+        this.repulsionZoneRadius = this.repulsionZoneRatio * this.pointSize;
+        // TODO init vertex degree, same as AbstractByEdgeNumberForce, maybe we should make that an interface with a default method instead of an abstract class
```
flo-dup: We need to remove this TODO before merging. Add a method in layoutContext to do this, it would avoid the duplication and make it more understandable by naming it properly.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/BoundingBox.java
```diff
@@ -5,6 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  * SPDX-License-Identifier: MPL-2.0
  */
+
```
flo-dup: No need to touch that file

diagram-util/src/main/java/com/powsybl/diagram/util/layout/postprocessing/NoPostProcessing.java
flo-dup: Delete this class and put instead a noOp static method in PostProcessing interface:
```java
static <V, E> PostProcessing<V, E> noOp() {
   return layoutContext -> { };
}
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/Layout.java
flo-dup: Avoid the redundant timing logic duplication by introducing a runAndMeasure method, something like
```java
private void runAndMeasure(String phaseName, Runnable phase) {
    long start = System.nanoTime();
    phase.run();
    long end = System.nanoTime();
    LOGGER.info("{} took {} s", phaseName, (end - start) / 1e9);
}
```

network-area-diagram/src/main/java/com/powsybl/nad/layout/Atlas2ForceLayout.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class Atlas2ForceLayout extends AbstractLayout {
+    private static final double SCALE_COEFFICIENT = 190.757;
+    private static final double SCALE_EXPONENT = -0.458;
+
+    private final Setup<Node, Edge> setup;
+    //maybe change the class name to not be confused with NAD LayoutParameters ?
```
flo-dup: you can remove this comment

network-area-diagram/src/main/java/com/powsybl/nad/layout/Atlas2ForceLayout.java
```diff
@@ -0,0 +1,108 @@
...
+                    new OverlapPreventionPostProcessingParameters.Builder().build()
+                )
+        );
+    }
+
+    @Override
+    protected void nodesLayout(Graph graph, LayoutParameters layoutParameters) {
+        LayoutContext<Node, Edge> layoutContext = new LayoutContext<>(graph.getJgraphtGraph(layoutParameters.isTextNodesForceLayout()));
+        double scale = SCALE_COEFFICIENT * Math.pow(layoutContext.getSimpleGraph().vertexSet().size(), SCALE_EXPONENT);
+        //TODO should we use the layoutParameters maxSteps to set Atlas2Parameters maxSteps ?
```
flo-dup: yes, and also for timeoutSeconds

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicForceLayout.java
```diff
@@ -34,6 +35,7 @@ public class BasicForceLayout extends AbstractLayout {
     private final boolean repulsionForceFromFixedPoints;
     private final boolean attractToCenterForce;
 
+    //TODO maybe deprecate this in favour of directly giving the BasicForceLayoutParameters, to be consistent with Atlas2ForceLayout
```
flo-dup: indeed, you can deprecate the constructor with booleans, but keep the empty constructor which takes all the parameters by default.

network-area-diagram/src/main/java/com/powsybl/nad/layout/LayoutParameters.java
flo-dup: Good question. This parameter could be seen as an expert parameter (whereas the timeoutSeconds is more understandable to the end-user) and thus only available through the various algorithms parameters. Let's keep that for another PR though.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
```diff
@@ -27,8 +29,8 @@ public final class BasicForceLayoutParameters {
...
+    private final boolean activateAttractToCenterForce;
```
flo-dup: Use the adjective rather than the verb (keep that for methods) and simplify.
Also in Atlas2Parameters.
```suggestion
    private final boolean repulsionFromFixedPointsEnabled;
    private final boolean attractToCenterEnabled;
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: Done in https://github.com/powsybl/powsybl-diagram/pull/721/commits/ef9fefade3795cdefb6d06ec42a4e5a6af1b6e32

diagram-util/src/main/java/com/powsybl/diagram/util/layout/postprocessing/NoPostProcessing.java
NathanDissoubray: If I do that, how will we give a PostProcessing that does nothing ? do we always call noOp ?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: done in 7d52ef7d9f1f

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: done in 8f68a8bf99bc

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractByEdgeNumberForce.java
NathanDissoubray: done in aa849b85e253

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/BoundingBox.java
NathanDissoubray: Indeed, done in 75b57d3f3c64

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceByEdgeNumberNoOverlapLinear.java
NathanDissoubray: it would be strange to put the force degree init inside layoutContext since layoutContext doesn't really have a direct link with force (since those are objects known to some LayoutAlgorithm, but not all layout algorithms need forces).
How about making an util class to be used by degree based forces ? Basically, we remove AbstratDegreeBasedForce.

The ideal case would be to be able to say that RepulsionForceDegreeBasedNoOverlapLinear (quite a mouthful), is both a AbstractNoOverlapForce and a AbstractDegreeBasedForce. This is not possible, so maybe Abstract Classes are not the good thing to do, which is why I said maybe interface with default methods ?

I think if we do an utility function for the init of degree based forces, then we also remove both AbstractDegreeBasedForce and AbstractNoOverlapForce, and we make an initDegree, and an initNoOverlap method to be used freely by force classes. It would be more coherent that way. And if we do this, we can remove the init method from the Force interface

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicForceLayout.java
NathanDissoubray: If I accept the BasicForceLayoutParameters as a parameter, then layoutParameters.maxSteps and layoutParameters.timeoutSeconds are redundant since those are already in the BasicForceLayoutParameters
And since I use a builder for this BasicForceLayoutParameters, I can't use those in the class constructor BasicForceLayout (unless I require layoutContext as a parameter).

But it seems strange to me to have in two places like that, I feel like it would be better to remove those parameters from the layoutContext

diagram-util/src/main/java/com/powsybl/diagram/util/layout/Layout.java
NathanDissoubray: done in b67cfcbce015

network-area-diagram/src/main/java/com/powsybl/nad/layout/Atlas2ForceLayout.java
NathanDissoubray: done in 56ab946c2343

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
NathanDissoubray: done in 59c8a255c843

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: we could convert the timeout to a deadline (so the deadline would be startTime + timeout) and give the deadline to run of the Setup, of the LayoutAlgorithm and of the PostProcessing
That would require changing the signature of those interface, which would be a breaking change. Unless we put the deadline inside the LayoutContext, but that would be strange

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: done in fd4733e08cf8

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: done in fd4733e08cf8

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
flo-dup: Ok, let's skip that in current PR, and support the timeoutSeconds in a separate PR. Indeed, we'd rather do directly the clean way with an `ExecutorService`,  `InterruptedException` and `Future.get(timeout)`.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
flo-dup: You forgoy to remove it from the documentation (atlas2Parameters.md + ieee_118 corresponding file)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: should we do that before or after #733 ?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: indeed, done with 0b35bc518c7a

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
flo-dup: You should remove Force keyword from those variable names as suggested to avoid such a long name

diagram-util/src/main/java/com/powsybl/diagram/util/layout/postprocessing/NoPostProcessing.java
flo-dup: how? simply by replacing `new NoPostProcessing<>()` with `PostProcessing.noOp()`.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceByEdgeNumberNoOverlapLinear.java
flo-dup: The degree init is just caching the degree, it has nothing related to force, so you could put it in LayoutContext under a cacheDegree method.

No, default in interface should rather not be used for providing features (default methods were first meant to avoid breaking changes when enriching an API and we've decided to keep it for that). And anyway the compiler will report a conflict if you have two interfaces with default methods.

You want to replace the init method in Force by two methods initDegree and initNoOverlap? Just try to think about the related documentation you'll need to write. You should not put methods specific to an implementation in an interface.

network-area-diagram/src/main/java/com/powsybl/nad/layout/BasicForceLayout.java
flo-dup: Yes they are redundant. If you want to avoid this redundancy and avoid checking their consistency when nodesLayout is called, then it would be better to remove thoses parameters from the BasicForceLayoutParameters (do not remove them from the LayoutParameters).

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/Force.java
```diff
@@ -20,6 +22,17 @@ default void init(LayoutContext<V, E> layoutContext) {
         // most forces do not need anything to init, default is empty
     }
 
+    /**
+     * Initializes all the forces using the layoutContext
+     * @param forces forces to be initialized
+     * @param layoutContext the context of the layout
+     */
+    static <V, E> void initAllForces(List<Force<V, E>> forces, LayoutContext<V, E> layoutContext) {
```
flo-dup: Please remove this, there's no gain in factorizing/hiding that tiny for loop

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractNoOverlapForce.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.diagram.util.layout.forces;
+
+import com.powsybl.diagram.util.layout.geometry.LayoutContext;
+
+/**
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public abstract class AbstractNoOverlapForce<V, E> implements Force<V, E> {
```
flo-dup: There's not much in that abstract class... Adding an heritage is not worth it. The content and heritage should be replaced, for instance with a record utility class (NoOverlapPointSize?) containing both double inputs and recalculating the pointSize with a method taking the layoutContext.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceByEdgeNumberNoOverlapLinear.java
NathanDissoubray: I see, indeed it would be strange to have those methods in the interface. I kept the init the in the Force interface since some forces still need init a bit differently
Should be fixed with 79b58623583e

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/BasicForceLayoutParameters.java
NathanDissoubray: indeed, I missed that, done with 6dac4e8d8996

diagram-util/src/main/java/com/powsybl/diagram/util/layout/postprocessing/NoPostProcessing.java
NathanDissoubray: done in 568cecef1728

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,225 @@
...
+        for (Point point : layoutContext.getMovingPoints().values()) {
+            previousForces.put(point, new Vector2D());
+            swingMap.put(point, 0.);
+        }
+        int i = 0;
+        int stoppingStep = layoutParameters.getMaxSteps();
+
+        while (i < stoppingStep) {
+            double graphSwing = 0.;
+            double newGraphSpeed;
```
flo-dup: define the variable when you calculate the value, no need to do that beforehand

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,225 @@
...
+                    layoutParameters.getSwingTolerance() * graphTraction / graphSwing,
+                    MAX_SPEED_DECREASE_RATIO * previousGraphSpeed,
+                    layoutParameters.getMaxGlobalSpeedIncreaseRatio() * previousGraphSpeed
+                    );
+            // calculate s(n) the speed of each node n
+            // store the forces on each node into the map of forces
+            // calculate D(n) the displacement of each node n
+            // reset forces on all points (we create a new vector2D so it won't affect forces in the map of forces)
+            updateAllPositions(layoutContext, newGraphSpeed, swingMap, previousForces);
+            if (isStable(newGraphSpeed, stoppingGlobalGraphSpeed)) {
```
flo-dup: The break is so close to the end of the loop... and leads to a code smell. Include it in the while condition to avoid sonar complaint.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedNoOverlapLinear.java
```diff
@@ -0,0 +1,93 @@
...
+    ) {
+        Vector2D force = Vector2D.calculateVectorBetweenPoints(otherPoint, point);
+        double magnitude = force.magnitude();
+        if (magnitude < repulsionZoneRadius) {
+            double forceIntensity;
+            if (magnitude < 2 * pointSize) {
+                forceIntensity = forceIntensityWithOverlap;
+            } else if (magnitude > 2 * pointSize) {
+                forceIntensity = forceIntensityNoOverlap / magnitude;
+            } else {
```
flo-dup: You can dismiss this else which has a close-to-zero probability to occur, and turn forceIntensity definition into a ternary expression.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedNoOverlapLinear.java
```diff
@@ -0,0 +1,93 @@
...
+        super.init(layoutContext);
+        this.repulsionZoneRadius = this.repulsionZoneRatio * this.pointSize;
+    }
+
+    @Override
+    public Vector2D apply(V vertex, Point point, LayoutContext<V, E> layoutContext) {
+        Vector2D resultingForce = new Vector2D();
+        int thisVertexDegree = point.getPointVertexDegree();
+        for (Map.Entry<V, Point> otherVertexPoint : layoutContext.getAllPoints().entrySet()) {
+            if (otherVertexPoint.getValue() != point) {
```
flo-dup, line 57: Rather test the key instead of the value.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
flo-dup: Rather after

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -0,0 +1,225 @@
...
+        while (i < stoppingStep) {
+            double graphSwing = 0.;
+            double newGraphSpeed;
+            double graphTraction = 0.;
+            //calculate forces
+            for (Map.Entry<V, Point> entry : layoutContext.getMovingPoints().entrySet()) {
+                Point point = entry.getValue();
+                for (Force<V, E> force : forces) {
+                    Vector2D resultingForce = force.apply(entry.getKey(), point, layoutContext);
+                    point.applyForce(resultingForce);
```
flo-dup, line 118: A more structural comment, which comes from the refactoring PR though; the need to call two `apply` methods to have the force applied on the point is strange.

Besides the fact that you give both the vertex and the point might lead to unexpected behaviour if the point does not correspond to the vertex. We would need to check the correspondance if we keep it that way. Rather send only the vertex to avoid the check.

The API would thus be something like
```java
void apply(V vertex, LayoutContext<V, E> layoutContext);
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/Force.java
NathanDissoubray: I don't really see the gain in having it duplicated at 3 different places either :thinking: is it because it's short that we don't care about the duplication ?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: I'll have to check, but I think passing both the vertex and the point was for a question of performance, but I'm not sure. I agree that it's strange to have to apply the force twice, this will need change

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractNoOverlapForce.java
NathanDissoubray: I still kept a `pointSize` variable inside both inheritor class, because with the record and the function taht calculates the point size, if we don't store that result somewhere, it means calculating it each time we use the force (which doesn't take that long, but still, it would be slower)

NathanDissoubray: Done in c676c8c6c798

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: good catch, I think it's like that because I used the graph speed somewhere inside the for loop that follows but still needed it after that (thus the definition because calculating it). But that's not needed anymore
Done in 611db756c7cd

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: I will have to define `newGraphSpeed` outside of the loop for that. On the first loop, I could initialize `newGraphSpeed` at double the stopping speed for example, (so that `isStable` doesn't say the result is stable).
That feels a bit strange to do

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedNoOverlapLinear.java
NathanDissoubray: that would change that line to `if (otherVertexPoint.getKey() != vertex)`
Could you explain why that's better than checking the value with the point ?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedNoOverlapLinear.java
NathanDissoubray: Done in 55e092ef77a4

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: Actually, most forces only need the point. Only forces that are applied on edges (Spring, EdgeAttractionLinear) need the vertex, to get the set of neighbors of the current vertex in the graph.
It might be a little bit slower to just give the vertex, because that means that each force will have to do a `layoutContext.getMovingPoints.get(vertex)` (whereas before we were iterating over the entries, and passing the point directly to the force). So that will likely triple the amount of get we do on the moving points.
I seem to remember that the get operations on the map where a non-negligeable part of the calculations (according to JProfiler).
I think it would be good to test the impact on performance of this change.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/Force.java
flo-dup: indeed, something which is that simple and short (in fact it's a one-liner here if you use forEach `forces.forEach(f -> f.init(layoutContext));`) does not need to be factorized.

NathanDissoubray: done in 5420d79f4de9

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractNoOverlapForce.java
flo-dup: I bet you a coffee you cannot prove me it's slower! 😋
Ok to store the value, but you should use a field in the record to hide that.
About the record, can't you remove the V, E by having `LayoutContext<?, ?>` as parameter? or by sending the graph size directly in factq

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/LayoutContext.java
```diff
@@ -29,34 +29,75 @@
...
             V target = graph.getEdgeTarget(edge);
+            // Remove loops
             if (source != target) {
                 locSimpleGraph.addEdge(graph.getEdgeSource(edge), graph.getEdgeTarget(edge));
             }
         }
         this.simpleGraph = locSimpleGraph;
     }
 
+    public void initDegree() {
```
flo-dup: You might call it twice if two degree-based forces are applied. Hence you probably need a boolean to tell that it's already cached.

Besides, just realized you kept initDegree instead of cacheDegree, don't you agree on calling it cache?

flo-dup:
> I think we should rename the Atlas2 parameter `attractToCenter` to not confuse it with `attractToCenterEnabled` With the name alone, it is not clear what it is. How about `attractToCenterIntensity` ? Didn't you ask me earlier to remove all the `intensity` from parameters name to keep it shorter ? I can't remember 🤔

No, I didn't, it was for `xxxForceEnabled` booleans. That parameter was `gravity` before. But yeah you're right, it's confusing! You already have an `attractToCenterIntensity` in Overlap, so that would be in line with that. You could have chosen `centerAttraction` also, which is similar to `edgeAttraction`.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/LayoutContext.java
NathanDissoubray: I probably just forgot to change the name, I don't really have a preference on the name, both init and cache are good, can change it to cache :+1: 
And indeed, it would be called as many times as there are forces that use a degree. Instead of a boolean, I could check if the value is null or not in order to cache, but that might be less clean

NathanDissoubray: nevermind, since the method is in layoutContext, I can just put a boolean in that

NathanDissoubray: done in de260c2a680a

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/AbstractNoOverlapForce.java
NathanDissoubray: With storing the point size, on two networks (number of substations, time in ms):
{548=1462, 5321=115922}
Without storing the point size and thus calculating it each time:
{548=1510, 5321=119229}
So yeah, not much difference, about 4s on the bigger network. I'll still keep it stored, because I find it cleaner. I'll make a field in the record

NathanDissoubray: done in 8a1bb66c4826

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: fixed the issue of the double break with 104ec1c77dad

flo-dup (approved):
Thanks for this big work! and nice to have that level of javadoc / rdt documentation

PR #724, opened 2025-08-21
NAD HVDC line edge type based on converter station

### LINKED ISSUE #723: [NAD] Tie LCC and VSC information to HDVC line BranchEdge

Currently the data of an HVDC line edge doesn't expose which type of converter station it is connected to 

A BranchEdge should specify if an HVDC line is to tied to a VCC or LCC converter station

### Describe the motivation

In order, in gridstudy, to be able to add a contextual menu to the line which opens a modification form of a given HVDC line and its converter stations we need this additionnal information

### Extra Information

Maybe the information would better fit the neighbouring node of the line and not the line itself, I'm not entirely familiar with the single-line-diagram-server and the diagram library

ayolab:

Fixes #723 

- [x] Yes
- [ ] No

branch edge type HvdcLineEdge removed.
users must handle HvdcLineLccEdge and HvdcLineVscEdge instead of the former generic type.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
             **/generated/**/*
         </sonar.exclusions>
 
-        <powsybl-core.version>6.8.1</powsybl-core.version>
+        <powsybl-core.version>7.0.0</powsybl-core.version>
 
     </properties>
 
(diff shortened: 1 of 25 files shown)


network-area-diagram/src/main/java/com/powsybl/nad/build/iidm/NetworkGraphBuilder.java
```diff
@@ -10,6 +10,7 @@
 import com.powsybl.diagram.util.IidmUtil;
```
alicecaron, line 10: can you describve more the sections in the pull request description? in particular the following:
- What kind of change does this PR introduce?
- What is the current behavior?
- What is the new behavior (if this is a feature change)?

ayolab: done

PR #730, opened 2025-09-16
Improve snakeline path in multi-substation SLD

NathanDissoubray:

Improve the algorithm that draws the snakeline between stations.

The snakelines tend to twist and turn without reason, or overlap which makes it less readable.

There is no more overlap. Snakelines turn only if there is actually an obstacle.

- [x] Yes
- [ ] No

com.powsybl.sld.layout.pathfinding.DijkstraPathFinder has been removed. Replace with com.powsybl.sld.layout.pathfinding.AStarPathFinder

**Other information**:

This gives better results but sadly it is also much slower (40~45s to run all SLD tests now). I'm working on a faster version, so we can merge this first or wait for the faster version and merge that directly. The branch with the faster version is on [faster_a_star](/powsybl/powsybl-diagram/tree/faster_a_star)

Here are some visual comparison between the previous path finding algorithm and the one implemented here:
Old:
image

New:
image

I also added a class AvailabilityGridImageDisplayed to help debug this kind of algorithm using a grid. It is not strictly necessary, but it can be helpful (and it helped me debug actually). We can keep it for future use or remove it.


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/pathfinding/PointHeading.java
```diff
@@ -0,0 +1,65 @@
...
+        return heading;
+    }
+
+    public void setHeading(PointInteger heading) {
+        this.heading = heading;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = hashCode;
```
flo-dup:
```suggestion
        return Objects.hash(point, heading);
```

NathanDissoubray:
Closing this, as altough the visual is better:
- it's too slow (40s to run all SLD test, against 8s currently), it would require another algorithm
- the quality of the code I wrote is subpar
It's simpler to start again from 0 rather than rewrite the code in this PR. Relevant information has been moved to the related issue #814

PR #733, opened 2025-09-29
Add Barnes-Hut optimization to Atlas2 NAD layout

NathanDissoubray:

#693 
Waiting for the merge of #721 

This PR adds a faster calculation of the repulsion force (known as Barnes-Hut) for Atlas2 in the NAD, which allows the bigger network like an 8k nodes to be done in less than 100s. Networks of less than 1000 nodes are done in less than 1s.
Those numbers do not include any potential post-processing (such as an overlap prevention, which might add 1 or 2 seconds of calculation on top of that).

Networks of 8k nodes take about 2k-3k seconds, networks of 1k nodes take about 8s

image
Some networks don't have as much as a gain as others, those are the networks where we can see a torsion in the graph (basically it forms an X), so Barnes-Hut can't optimize as much because nodes are in a less favorable position). We still gain time for all graph with this method though.
The visual does not change in any noticeable way.

- [ ] Yes
- [x] No

**Other information**:


diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
```diff
@@ -22,6 +22,10 @@ public final class Atlas2Parameters {
     private static final boolean DEFAULT_ACTIVATE_REPULSION_FORCE_FROM_FIXED_POINTS = true;
     private static final boolean DEFAULT_ACTIVATE_ATTRACT_TO_CENTER_FORCE = true;
     private static final double DEFAULT_ITERATION_NUMBER_INCREASE_PERCENT = 0;
+    private static final double DEFAULT_BARNES_HUT_THETA = 1.2;
+    /// See "It Pays to Be Lazy: Reusing Force Approximations to Compute Better Graph Layouts Faster"
```
flo-dup: Put this comment in the javadoc style

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/quadtreeupdateschedule/ConstantSchedule.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) 2025, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+package com.powsybl.diagram.util.layout.algorithms.quadtreeupdateschedule;
+
+/// See "It Pays to Be Lazy: Reusing Force Approximations to Compute Better Graph Layouts Faster"
```
flo-dup: include this comment in the javadoc one below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/QuadtreeContainer.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.powsybl.diagram.util.layout.geometry;
+
+/**
+ * A class that contains a quadtree. This is used for initialization purposes. When starting some algorithms, we create forces
+ * at the start that require a quadtree object, but we don't actually have the points objects needed to create the quadtree. Instead
+ * we use this to say we have a quadtree, and we will later actually create and use the quadtree
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class QuadtreeContainer {
```
flo-dup: Use Ref<> and RefObj<> from poswybl-commons instead

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -162,11 +160,44 @@ public void run(LayoutContext<V, E> layoutContext) {
...
+     */
+    private Collection<Point> getInteractingPoints(LayoutContext<V, E> layoutContext) {
+        if (layoutParameters.isActivateRepulsionForceFromFixedPoints()) {
+            return layoutContext.getAllPoints().values();
+        } else {
+            return layoutContext.getMovingPoints().values();
+        }
+    }
+
+  /**
```
flo-dup: Fix indentation

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * A quadtree is a structure that recursively divides a space in 4 until only a given number of points reside in each subdivided area. In this case,
+ * the quadtree will only contain 1 point in each leaf node.
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class Quadtree {
+    public static class QuadtreeNode {
+        // package private
+        short[][] childrenNodeId = {
```
flo-dup: use the flatten version directly, with a tiny comment, that would simplify the code

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+     *                       then the splitPredicate must return the same boolean for both
+     * @return the splitIndex at which the predicate changes from true to false on the sorted array.
+     * for all i in [startIndex, splitIndex[, predicate(points[i]) is true
+     * for all i in [splitIndex, endIndex[, predicate(points[i]) is false
+     * no guarantee is given for points outside the [startIndex, endIndex[ range
+     */
+    private short partitionPoints(Point[] points, short startIndex, short endIndex, Predicate<Vector2D> splitPredicate) {
+        // directly find the index such that the predicate is false, no need to sort the start of the array if it's already ok
+        short firstFalseIndex = startIndex;
+        while (firstFalseIndex < endIndex) {
```
flo-dup:
```suggestion
        while (firstFalseIndex < endIndex && splitPredicate.test(points[firstFalseIndex].getPosition())) {
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+            if (splitPredicate.test(points[firstFalseIndex].getPosition())) {
+                ++firstFalseIndex;
+            } else {
+                break;
+            }
+        }
+        if (firstFalseIndex == endIndex) {
+            return firstFalseIndex;
```
flo-dup: remove this, it is covered by the lines below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+ */
+package com.powsybl.diagram.util.layout.geometry;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.function.ToDoubleFunction;
+
+/// This quadtree implementation is a Java version of a C++ quadtree presented by Nikita Lisitsa on his lisyarus blog
```
flo-dup: Add that comment in the javadoc section below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: wouldn't it be a problem to use a short if a big graph like 50k nodes is used?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+        for (short index : barycenterIndex) {
+            // index is only -1 in the case of nothing being there
+            if (index != NO_CHILDREN) {
+                // get the mass / position of each quadrant and do a weighted sum (quite literally)
+                Point quadrantBarycenter = barycentersList.get(index);
+                Vector2D quadrantBarycenterPosition = new Vector2D(quadrantBarycenter.getPosition());
+                // do not use the massGetter, that's only for leaf nodes, quadrants contain leaf which already have their correct mass set
+                double quadrantMass = quadrantBarycenter.getMass();
+                quadrantBarycenterPosition.multiplyBy(quadrantMass);
+                barycenterPosition.add(quadrantBarycenterPosition);
```
flo-dup: To avoid creating many Vector2D:
```suggestion
                barycenterPosition.addScaled(quadrantBarycenter.getPosition(), quadrantBarycenter.getMass());
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceByEdgeNumberLinearBarnesHut.java
```diff
@@ -0,0 +1,126 @@
...
+            double nodeWidth,
+            List<Point> pointsToInteractWith
+    ) {
+        Quadtree quadtree = quadtreeContainer.getQuadtree();
+        Point barycenter = quadtree.getBarycenters()[nodeIndex];
+        // Two tests:
+        // Check the theta parameter ie width / distance < theta
+        // If a point is alone in a quadrant, but is not a leaf node, we could continue descending until the leaf node (or until the first condition is met), but this is not useful
+        // this second check only works if the point doesn't have any edges (because otherwise in Atlas2 its mass will be more than the default mass for example)
+        if (nodeWidth < barnesHutTheta * point.distanceTo(barycenter) || barycenter.getMass() == Point.DEFAULT_MASS) {
```
flo-dup: You are trying not to descend to the children of a leaf node with the second condition, as you mentioned the safeguard below. But a leaf node will have a barycenter with a mass greater than one. You should either remove the second test and keep the "safeguard" or change it to something like `quadtree.getNodes()[nodeIndex].isLeaf()`

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
flo-dup: Format this comment in the javadoc style

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
```diff
@@ -208,6 +220,39 @@ public Builder withIterationNumberIncreasePercent(double iterationNumberIncrease
...
+         */
+        public Builder withQuadtreeCalculationIncrement(int quadtreeCalculationIncrement) {
+            if (quadtreeCalculationIncrement <= 0) {
+                throw new IllegalArgumentException("The increment for the constant schedule has to be strictly positive");
+            }
+            this.quadtreeCalculationIncrement = quadtreeCalculationIncrement;
+            return this;
+        }
+
+        /// By default, the Barnes-Hut optimisation is activated with theta = 1.2
```
flo-dup: This interesting comment will only be seen by contributors of powsybl, you'd better add this in a javadoc comment

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/quadtreeupdateschedule/ConstantSchedule.java
flo-dup: include this comment in the javadoc one below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -108,10 +100,16 @@ public void run(LayoutContext<V, E> layoutContext) {
         int stoppingStep = layoutParameters.getMaxSteps();
         boolean changedStoppingStep = false;
 
+        ConstantSchedule quadtreeUpdateSchedule = new ConstantSchedule(layoutParameters.getQuadtreeCalculationIncrement());
+
         while (i < stoppingStep) {
             double graphSwing = 0.;
             double newGraphSpeed;
             double graphTraction = 0.;
+            if (quadtreeUpdateSchedule.isTimeToUpdate(i) && layoutParameters.getBarnesHutTheta() > 0) {
```
flo-dup: invert conditions to check first if BarnesHut is enabled before checking the schedule

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
```diff
@@ -268,5 +315,13 @@ public boolean isActivateAttractToCenterForce() {
     public double getIterationNumberIncreasePercent() {
         return iterationNumberIncreasePercent;
     }
+
+    public double getBarnesHutTheta() {
```
flo-dup, line 281: Add also a `isBarnesHutEnabled` convenient method

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
```diff
@@ -208,6 +220,39 @@ public Builder withIterationNumberIncreasePercent(double iterationNumberIncrease
             return this;
         }
 
+        /**
+         * The theta parameter used in the Barnes-Hut approximation. The bigger the theta, the more aggressive the optimization will be,
+         * but that might lead to less visual quality. A bigger value will also generally reduce runtime. Default is {@value DEFAULT_BARNES_HUT_THETA}
+         * @param barnesHutTheta the theta for the barnes-hut optimization
+         * @return  the instance of this Builder with the `barnesHutTheta` changed
+         */
+        public Builder withBarnesHutTheta(double barnesHutTheta) {
```
flo-dup, line 195: Add a withBarnesHutDisabled convenient method which sets it to 0.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -151,7 +149,7 @@ public void run(LayoutContext<V, E> layoutContext) {
             if (!changedStoppingStep && isStable(newGraphSpeed, stoppingGlobalGraphSpeed)) {
                 stoppingStep = Math.min(
                         layoutParameters.getMaxSteps(),
-                        (int) (i * (1 + layoutParameters.getIterationNumberIncreasePercent() / 100))
+                        (int) (i * (1 + layoutParameters.getIterationNumberIncreasePercent() / 100) * (1 + layoutParameters.getBarnesHutTheta() / 8))
```
flo-dup: Like what I said in Atlas2 PR, remove that post-stability stoppingStep increase. If the isStable logic is not enough with Barnes-Hut, we should rather change that stopping logic instead of that steps increase.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
flo-dup: Fix indentation

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
```diff
@@ -162,11 +160,44 @@ public void run(LayoutContext<V, E> layoutContext) {
...
+        }
+    }
+
+  /**
      * Calculate the swing of the point, as the magnitude of the difference between the previous and the current force applied to the point
      * @param point the point the force is applied to
      * @param previousForce the force that was previously applied on the point
      * @return the swing of the point
      */
+
```
flo-dup:
```suggestion
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinearBarnesHut.java
```diff
@@ -0,0 +1,126 @@
...
+            double nodeWidth,
+            List<Point> pointsToInteractWith
+    ) {
+        Quadtree quadtree = quadtreeContainer.getQuadtree();
+        Point barycenter = quadtree.getBarycenters()[nodeIndex];
+        // Two tests:
+        // Check the theta parameter ie width / distance < theta
+        // If a point is alone in a quadrant, but is not a leaf node, we could continue descending until the leaf node (or until the first condition is met), but this is not useful
+        // this second check only works if the point doesn't have any edges (because otherwise in Atlas2 its mass will be more than the default mass for example)
+        if (nodeWidth < barnesHutTheta * point.distanceTo(barycenter) || barycenter.getMass() == Point.DEFAULT_MASS) {
```
flo-dup: It seems you are trying not to descend to the children of a leaf node with the second condition (see the "safeguard" comment below). But a leaf node will often have a barycenter with a mass greater than one. You should either remove the second test and keep the "safeguard" or change it to something like `quadtree.getNodes()[nodeIndex].isLeaf()`

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: Add that comment in the javadoc section below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: Use four final parameters instead of a 2D array, that would make the code clearer.
As the parameters will be final, create a constructor and create two static method, one for creating a leaf, one for creating an internal node.

flo-dup: I think using a `short` will become a problem on big graphs like 50k nodes. Try to use this on [Synthetic USA](https://github.com/MATPOWER/matpower/blob/master/data/case_SyntheticUSA.m) case for instance

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup:
```suggestion
        while (firstFalseIndex < endIndex && splitPredicate.test(points[firstFalseIndex].getPosition())) {
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: remove this, it is covered by the lines below

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: To avoid creating many many Vector2D:
```suggestion
                barycenterPosition.addScaled(quadrantBarycenter.getPosition(), quadrantBarycenter.getMass());
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+public class Quadtree {
+    public static class QuadtreeNode {
+        // package private
+        short[][] childrenNodeId = {
+                {NO_CHILDREN, NO_CHILDREN},
+                {NO_CHILDREN, NO_CHILDREN}
+        };
+
+        public short[] getChildrenNodeIdFlatten() {
+            return new short[] {
```
flo-dup: `return new short[0]` if it is a leaf.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+        if (firstIndex == lastIndex) {
+            // no children
+            return NO_CHILDREN;
+        }
+        short newNodeIndex = (short) nodesList.size();
+        nodesList.add(new QuadtreeNode());
+        Point nodeBarycenter = new Point(0, 0);
+        barycentersList.add(nodeBarycenter);
+
+        if (firstIndex + 1 == lastIndex) {
```
flo-dup: Group the two ifs together

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,312 @@
...
+            int remainingDepth
+    ) {
+        if (firstIndex == lastIndex) {
+            // no children
+            return NO_CHILDREN;
+        }
+        short newNodeIndex = (short) nodesList.size();
+        nodesList.add(new QuadtreeNode());
+        Point nodeBarycenter = new Point(0, 0);
+        barycentersList.add(nodeBarycenter);
```
flo-dup: Add the QuadtreeNode to the list once they're fully constructed. Same for nodeBarycenter and barycentersList. Avoid "half-baked" states; adding a first version, then getting it and modifying the value is confusing

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/QuadtreeContainer.java
flo-dup: Use Ref<> and RefObj<> from poswybl-commons instead

flo-dup (changes requested):

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: Done in 15faef2ee527

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: moved this comment to the `withBarnesHutTheta` and `withBarnesHutDisabled` in 15faef2ee527 and cf59a8f8e760

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/quadtreeupdateschedule/ConstantSchedule.java
NathanDissoubray: Done in 15faef2ee527

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: Done in efa9668e22f0

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: Done in cf59a8f8e760, also used it in e8c28402765f

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: Done in cf59a8f8e760

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: This was removed during the merge since the `iterationNumberIncreasePercent` was removed during the Atlas2 PR. However, I think it would be good to have a bit more strict of a stopping condition the higher the value of the theta is (since we are approximating the forces on each point with Barnes-Hut, we are also approximating the metric we use for stopping)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: Done in 1939caf471ff

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/Atlas2ForceLayoutAlgorithm.java
NathanDissoubray: This seems to have been removed in the merge or in the previous Atlas2 PR

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in cae85b915f67

diagram-util/src/main/java/com/powsybl/diagram/util/layout/algorithms/parameters/Atlas2Parameters.java
NathanDissoubray: Nevermind, the correct version of the function is in cbf46e8ce303

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinearBarnesHut.java
NathanDissoubray: I think I'll just keep the safeguard below, since the `isLeaf` method would basically do what is done below in the function, ie checking if the index corresponds to `Quadtree.NO_CHILDREN`, but I would have to get the list of index that are actual children to continue with the recursion with `generatePointInteractionList`

NathanDissoubray: Done in 866ce4afa19f

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/QuadtreeContainer.java
NathanDissoubray: Done in b582399e8332

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in d24e3ca662fa

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Do you mean one parameter for each quadrant ? like top left, top right, bottom left, bottom right ?
In that case, they can't be final, or else I can't modify the index of children node.
So I should use a static method that takes 4 index to build an internal node ?

NathanDissoubray: well, apart from the fact that 63k nodes seems to be more than what this code can handle (see image below), the code does not crash.
But it's true that in theory, using a short, we shouldn't be able to build the quadtree (since at most we can have 32 767 values in the tree)
image
I can replace this by an int and that would definitely solve the problem for any network.

NathanDissoubray: it's interesting, when using a short or an int, it doesn't take the same number of steps to finish. Maybe there was some overwriting going on inside the quadtree :thinking:

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: If I keep the way we choose an index, I cannot do that. If I don't add the quadtree node to the list from the start, we can have the following case:
- let's say newNodeIndex is 4 (there are 4 nodes already, I would like to add a new one at index 4)
- I don't add a node the node to the list, the list is still of size 4
- I build the quadtree for one quadrant of my current node (which I want to put in the list at index 4), I look at the list, there are only 4 nodes, my new node for this quadrant will be at index 4

We are in a situation where two different nodes will be at the same index, which we don't want.

If I only get the index before adding to the list, I'm getting out of bounds errors, which i'm not yet sure why, I'll keep searching.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in d386a445eab4

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: done in 02e92f8dedc1

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: done in 964fe6c627c0

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: done in 083488173ce6

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: to know that a node is a leaf, I will need to check that all its childrenNodeId are equal to `NO_CHILDREN`
If that's not the case, it means that I did the work in double by iterating, since whatever called this will probably iterate over the int[] as well

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: By the way, here is what it looks like after 50k iterations and 1h50 of calculations
image
It's not there yet but we can see a shape somewhat starting. Well, this code is not really suited for networks that big it seems.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinearBarnesHut.java
NathanDissoubray: I just removed the comments in the first commit, the actual removal of the condition is in 8d8741dae134

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: fixed the issue by having `getChildrenNodeIdFlatten` only return index that correspond to a child directly, in 9555f2af57b6

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Note quite what you wanted, but I moved the node creation closer to where it's actually modified in 848a70191f35
There is still an "half-baked" state, but it's for a shorter time now.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/QuadtreeContainer.java
NathanDissoubray: Modified some things in cec00de918ea

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Vector2D.java
flo-dup: The created object here is not used and therefore trashed just after. I think you wanted to write
`this(0, 0);`

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+
+        QuadtreeNode(Point nodeBarycenter) {
+            this.nodeBarycenter = nodeBarycenter;
+        }
+
+        /**
+         * @return all the index of children that actually exists (ie all quadrants that have at least a point in them)
+         */
+        public int[] getRealChildrenNodeIndex() {
+            return Stream.of(
```
flo-dup: IntStream.of instead to avoid boxing ints and mapToInt

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+ * A quadtree is a structure that recursively divides a space in 4 until only a given number of points reside in each subdivided area. In this case,
+ * the quadtree will only contain 1 point in each leaf node.
+ * <p>This quadtree implementation is a Java version of a C++ quadtree presented by Nikita Lisitsa on his lisyarus blog.
+ * The article is called "<a href="https://lisyarus.github.io/blog/posts/building-a-quadtree.html">Building a quadtree in 22 lines of code</a>"</p>
+ * @author Nathan Dissoubray {@literal <nathan.dissoubray at rte-france.com>}
+ */
+public class Quadtree {
+    public static class QuadtreeNode {
+        // package private
+        int[][] childrenNodeIndex = {
```
flo-dup: To make it more explicit, replace with
```java
public int bottomLeft;
public int bottomRight;
public int topLeft;
public int topRight;
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+        if (
+            // if there is only a single Point in this node, it's a leaf, barycenter is the point
+                firstIndex + 1 == lastIndex
+                || remainingDepth == 0
+            // if the size of the points in an area did not diminish, it might be because two or more points are at the same position, or very close
+            // use the first two index checks to not have to calculate the equality between all points in the range each time
+                || firstIndex == previousFirstIndex && lastIndex == previousLastIndex && checkPointPositionEquality(points, firstIndex, lastIndex)
+        ) {
+            int newNodeIndex = nodesList.size();
+            Point nodeBarycenter = new Point(0, 0);
```
flo-dup: No need to create it at 0,0 first. You can now directly create the correct position: 
```java
Point nodeBarycenter = createLeafBarycenter(points, firstIndex, lastIndex, massGetter);
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+        int xUpperSplitIndex = partitionPoints(points, ySplitIndex, lastIndex, isLeft);
+
+        BoundingBox bottomLeftBb = new BoundingBox(boundingBox.left(), boundingBoxCenter.getY(), boundingBoxCenter.getX(), boundingBox.bottom());
+        BoundingBox bottomRightBb = new BoundingBox(boundingBoxCenter.getX(), boundingBoxCenter.getY(), boundingBox.right(), boundingBox.bottom());
+        BoundingBox topLeftBb = new BoundingBox(boundingBox.left(), boundingBox.top(), boundingBoxCenter.getX(), boundingBoxCenter.getY());
+        BoundingBox topRightBb = new BoundingBox(boundingBoxCenter.getX(), boundingBox.top(), boundingBox.right(), boundingBoxCenter.getY());
+
+        int newNodeIndex = nodesList.size();
+        //need to add it first here even though we modify it just after, because buildQuadtree is recursive. Need to conserve order
+        Point nodeBarycenter = new Point(0, 0);
```
flo-dup: and same here `nodesList.add(new QuadTreeNode())` (let QuadTreeNode deal with the missing barycenter and later below `quadTreeNode.nodeBarycenter = createNodeBarycenter(nodesList, nodesList.get(newNodeIndex))`)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+        nodesList.add(new QuadtreeNode(nodeBarycenter));
+
+        nodesList.get(newNodeIndex).childrenNodeIndex[0][0] = buildQuadtree(nodesList, points, bottomLeftBb, firstIndex, xLowerSplitIndex, massGetter, firstIndex, lastIndex, remainingDepth - 1);
```
flo-dup: Avoid the 4 gets by introducing a variable for the added QuadTreeNode. And then 
```java
quadTreeNode.bottomLeftIndex = buildQuadTree(...);
quadTreeNode.bottomRightIndex = buildQuadTree(...);
...
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+    private int partitionPoints(Point[] points, int startIndex, int endIndex, Predicate<Vector2D> splitPredicate) {
+        // directly find the index such that the predicate is false, no need to sort the start of the array if it's already ok
+        int firstFalseIndex = startIndex;
+        while (firstFalseIndex < endIndex && splitPredicate.test(points[firstFalseIndex].getPosition())) {
+            ++firstFalseIndex;
+        }
+
+        for (int i = firstFalseIndex + 1; i < endIndex; ++i) {
+            if (splitPredicate.test(points[i].getPosition())) {
+                // swap points
```
flo-dup, line 205: Use `Collections::swap` instead

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+     * @param splitPredicate a function that returns either true or false given a Vector,
+     *                       it must be consistent ie if two vectors are equals with regards to .equals(),
+     *                       then the splitPredicate must return the same boolean for both
+     * @return the splitIndex at which the predicate changes from true to false on the sorted array.
+     * for all i in [startIndex, splitIndex[, predicate(points[i]) is true
+     * for all i in [splitIndex, endIndex[, predicate(points[i]) is false
+     * no guarantee is given for points outside the [startIndex, endIndex[ range
+     */
+    private int partitionPoints(Point[] points, int startIndex, int endIndex, Predicate<Vector2D> splitPredicate) {
+        // directly find the index such that the predicate is false, no need to sort the start of the array if it's already ok
```
flo-dup, line 197: this is an optimization which produces extra code for a quite low probability case, shouldn't we rather remove this?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+     * @param lastIndex the index of the last point contained in the bounding box (excluded)
+     * @param massGetter the function that gives the mass of a point
+     * @param previousFirstIndex the previous index of the first point contained in the previous bounding box (included).
+     *                           Used for an optimisation to check if two or more points are at the same position (to avoid recursively dividing until the depth limit)
+     * @param previousLastIndex the previous index of the last point contained in the previous bounding box (excluded).
+     *                           Used for an optimisation to check if two or more points are at the same position (to avoid recursively dividing until the depth limit)
+     * @param remainingDepth how many more division of the space can we make for this given bounding box
+     * @return the index of the most parent node (if the depth is 0, this is the root of the quadtree, if the depth is 1, this is the index of a child of the root node, etc...)
+     */
+    private int buildQuadtree(
```
flo-dup, line 121: We could have the following signature which makes it more understandable and avoids passing indices through all the recursive calls. See suggestion here: b8aa6dcbbdb3576eedc02b759da3fd21272f692d 
```java
private int buildQuadtree(
            List<QuadtreeNode> nodesList,
            List<Point> points,
            BoundingBox boundingBox,
            ToDoubleFunction<Point> massGetter,
            int previousSize,
            int remainingDepth
```

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * Build a quadtree with the given points and a function to get the mass of the point
+     * @param points the list of points we want a quadtree for
+     * @param massGetter a function that associates each point to a mass, used to calculate barycenter. This is used if we want
+     *                   to calculate barycenter differently. For example, we might want to give all the points a mass of 1, or we might
+     *                   want each point to have a mass equal to its number of edges + 1; this allows flexibility
+     */
+    public Quadtree(Collection<Point> points, ToDoubleFunction<Point> massGetter) {
+        bb = BoundingBox.computeBoundingBox(points);
+        List<QuadtreeNode> nodesList = new ArrayList<>();
```
flo-dup: We expect at least points.size() elements in that list, don't we? hence to avoid resizing the list a lot just pass that size to the `ArrayList` constructor

NathanDissoubray: yes, an even better lower bound would be:
$$p = points.size()$$
$$n = \lfloor \log_4(p) \rfloor$$
$$\sum_{i=0}^n \frac{p}{4^i}$$
since the best tree you can make is one where all the leafs are at the same level (p/4^0), each group of 4 leafs has 1 parent (so add p/4^1), each one of those has a parent too (p/(4^2)) and the number of levels would be the log in base 4 of the number of points (rounded down)

NathanDissoubray: which is that value according to WolframAlpha (it's just the sum of a geometric series)
image
But that might be going a bit overboard to calculate a lower bound :laughing:

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: In the article I used it was mentionned:
"
Also note that I'm explicitly repeating tree.nodes[result] instead of storing it in a local reference node & current = tree.nodes[result], because the build_impl function adds new nodes to the nodes array, and the reference might get invalidated.
"
Where `build_impl` would be `buildQuadtree` for us.
I think it would be true for C++ since it's a pointer, and memory re-allocation in `build_impl` would invalidate the reference (if the array is moved, the reference is incorrect). But I don't think this is an issue in java, since the object will still exist even if the array is moved. I'll change it and see what happens

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: `Collections::swap` works on List<?> not on Array, so I need to manually swap the elements as there is no Arrays::swap that exists (and I don't think converting this to a list just to do swaps would be good for performance)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: I don't think it's that uncommon, if you have a cluster of points it's possible that they all are in the same quadrant of the current node already

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Vector2D.java
NathanDissoubray: Indeed, done in 5de8c620999c

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: I did not know of this method, thanks, done in fd17d090f5d7

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: done in 59342ec02af1

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in 518915fa3f53, there was no issue

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in 10dea7aebff5
Tell me if you would prefer the "better" lower bound (it's closer, but in both cases we will need to resize anyway, so the "better" option will probably just do more calculations for the same effective result)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Using subList would provide the same functionnal result but might be slower. I'll do some testing to see the impact, if it's not too significant it will be better to have this shorter signature without indices

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: After seeing https://github.com/powsybl/powsybl-diagram/commit/b8aa6dcbbdb3576eedc02b759da3fd21272f692d I guess if we are using a list we can indeed using `Collections::swap`

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Format: {number of substations=[times in ms]}

Times with sublist: {548=[767, 347, 307], 5321=[57940, 58517, 57926]}
Times without sublist: {548=[800, 316, 299], 5321=[60149, 61187, 59879]}
So it's equivalent, even slightly faster with sublist, and the code is much cleaner, thank you for the suggestion @flo-dup ! I did not know about sublist.

I think since it's a view of the same list and we are not creating a new array each time, that's why it's a bit faster.

Done in 33ac1851da7f

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Actually, it wouldn't even be correct if we remove that part, consider the following (for simplicity, I'll just write the result of the test of the predicate on the points):
`[true, false, false, true]`
Start at i = 1; false, don't swap
i = 2; false, don't swap;
i = 3; true, swap with firstFalseIndex = 0; we get the same list as the beginning, we didn't actually partition the points.

To fix this, let's say we check that the element at firstFalseIndex is false, if that's not the case, we will start swapping at the next index. But then, you will have the same issue with:
`[true, true, false, true]`

To fix this, we have no choice but to find the first index that is false, thus the little loop to do so.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: When doing that, I found a subtle bug. Before change:
```java
private void setLeafBarycenter(List<Point> points, Point nodeBarycenter, ToDoubleFunction<Point> massGetter) {
        Point leafPoint = points.getFirst();
        nodeBarycenter.setPosition(leafPoint.getPosition());
        double totalMass = massGetter.applyAsDouble(leafPoint);
        for (int i = 1; i < points.size(); ++i) {
            // Springy and Atlas define mass differently for forces
            // for springy it's the literal mass, for Atlas it's the vertex degree + 1, just use a function to accommodate everyone
            totalMass += massGetter.applyAsDouble(points.get(i));
        }
        nodeBarycenter.setMass(totalMass);
    }
```

after change:
```java
private Point createLeafBarycenter(List<Point> points, ToDoubleFunction<Point> massGetter) {
        Point leafPoint = points.getFirst();
        Point nodeBarycenter = new Point(leafPoint.getPosition().getX(), leafPoint.getPosition().getY());
        double totalMass = massGetter.applyAsDouble(leafPoint);
        for (int i = 1; i < points.size(); ++i) {
            // Springy and Atlas define mass differently for forces
            // for springy it's the literal mass, for Atlas it's the vertex degree + 1, just use a function to accommodate everyone
            totalMass += massGetter.applyAsDouble(points.get(i));
        }
        nodeBarycenter.setMass(totalMass);
        return nodeBarycenter;
    }
```
The important difference is that in `setLeafBarycenter` we use `setPosition`, thus using the same position object for the barycenter of the node and the the actual point. In `createLeafBarycenter`, it's a separate object, since we create the position of the barycenter by getting individually the x and y coordinates of the leaf node.

That alone wouldn't cause a bug, but since we don't calculate the quadtree at each iteration of Atlas2, in the old version the quadtree leaf node would get "updated" because the point in itself would move, even though we didn't re-calculate the quadtree.
In the new version, the quadtree stays the same until we calculate it again. This gives a difference in the look of the result:

With `setLeafBarycenter`:
image

With `createLeafBarycenter`:
image

Another example:
`setLeafBarycenter`
image

`createLeafBarycenter`
image

I know the updating of the quadtree through having the same vector object with `setPosition` was unintended and technically a bug. But I find that version more readable. Could be a case of "not a bug, it's a feature" :laughing: 
What should we do with this ?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: We can approximate this lower bound as `4/3*p`
image

And since it's a lower bound, we know that the list will need resizing anyway, so we can start at `2*4/3*p`

flo-dup: You could choose `2p` instead, as there are some trees which will have their size between that lower bound and `2p`. Isn't  `4p` an upper bound by the way?

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: I think we could keep one loop only, but at the cost of doing some useless swaps (`i <-> i`), leading to a less understandable code. Let's keep it like you did.

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: I think the upper bound depends on the `MAX_RECURSION_DEPTH` as well. 2p might work in most cases, I'll try that

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in c3fcdd1c1d5b

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Done in 7de4171ebea9
Maybe I can remove the explanation in the comment on the lower bound

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: Avoiding the "half-baked" state would mean
- either insert a null in the ArrayList before adding the proper QuadtreeNode; then it's the list which is in "half-baked" state!
- or completely reverse the indexing (root at the end instead), adding the parent after its children by simply moving the add. This works like a charm and might be considered as the code might be more understandable. Nonetheless the resulting array might be a bit confusing: traversing the quadtree means starting from the last index and then follow the indices (which always point backward then)

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: Ok, let's keep it. Please add doc about this!

flo-dup: Don't forget to replace the setLeafBarycenter method by a createLeafBarycenter in any case. The asked refactoring is not directly related to this bug/feature.

NathanDissoubray: Yes indeed, this is what was done in c50e2e693

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
NathanDissoubray: Should be changed with c50e2e693

diagram-util/src/main/java/com/powsybl/diagram/util/layout/geometry/Quadtree.java
flo-dup: Ok, all this already in the recent commits, sorry!

flo-dup (approved):

PR #743, opened 2025-10-20
change attributes' visibility in DefaultLabelProvider

carojeandat:

No

Feature

The class attributes of DefaultLabelProvider are private

The class attributes are now protected, allowing subclasses to access and use them when overriding DefaultLabelProvider methods.

- [ ] Yes
- [x] No

**Other information**:


EstherDarkish (dismissed):

flo-dup (changes requested):
Could you rather use protected getters and keep the fields private? It would be cleaner, more flexible, and it keeps the internal structure hidden

PR #747, opened 2025-10-24
Align the line's names in the Single Line Diagrams.

EstherDarkish:

Align the line names in Single Line Diagrams.

When a line is ended with an icon, its name is not on the same horizontal line as another line without an icon. This produces overlapping text, notably when using diagonal names.

Line names are aligned.

- [ ] Yes
- [x] No

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/test/resources/Test3WTFeederInfoArrowVoltageLevelWithAllPossibleFeederInfos.svg
+++ b/single-line-diagram/single-line-diagram-core/src/test/resources/Test3WTFeederInfoArrowVoltageLevelWithAllPossibleFeederInfos.svg
@@ -182,10 +182,10 @@
                     <circle class="sld-winding" cx="8" cy="19" r="5"/>
                 </g>
                 <g class="sld-top-feeder" id="id3WT_95_TWO" transform="translate(115.0,80.0)">
-                    <text class="sld-label" id="id3WT_95_TWO_95_N_95_LABEL" x="-5.0" y="-5.0">3WT</text>
+                    <text class="sld-label" id="id3WT_95_TWO_95_N_95_LABEL" x="-5.0" y="-11.0">3WT</text>
                 </g>
                 <g class="sld-top-feeder" id="id3WT_95_THREE" transform="translate(165.0,80.0)">
-                    <text class="sld-label" id="id3WT_95_THREE_95_N_95_LABEL" x="-5.0" y="-5.0">3WT</text>
+                    <text class="sld-label" id="id3WT_95_THREE_95_N_95_LABEL" x="-5.0" y="-11.0">3WT</text>
                 </g>
             </g>
         </g>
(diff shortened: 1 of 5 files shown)


carojeandat (dismissed):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/AbstractLabelProvider.java
```diff
@@ -105,9 +106,7 @@ protected LabelPosition getLabelPosition(Node node, Direction direction) {
         String positionName = "";
         double angle = 0;
         if (direction != UNDEFINED) {
-            yShift = direction == TOP
-                    ? -LABEL_OFFSET
-                    : ((int) (componentLibrary.getSize(node.getComponentType()).getHeight()) + LABEL_OFFSET);
+            yShift = (direction == TOP ? -LABEL_PADDING : LABEL_PADDING) + (componentLibrary.getSize(node.getComponentType()).getHeight() / 2);
```
flo-dup: This only works if the feeder component has a size `< 2 * LABEL_PADDING`. Any idea how to avoid this problem?

flo-dup: Why don't you make use of layoutParameters.getMaxComponentHeight()?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/AbstractLabelProvider.java
```diff
@@ -105,9 +105,7 @@ protected LabelPosition getLabelPosition(Node node, Direction direction) {
         String positionName = "";
         double angle = 0;
         if (direction != UNDEFINED) {
-            yShift = direction == TOP
-                    ? -LABEL_OFFSET
-                    : ((int) (componentLibrary.getSize(node.getComponentType()).getHeight()) + LABEL_OFFSET);
+            yShift = (direction == TOP ? -layoutParameters.getMaxComponentHeight() : layoutParameters.getMaxComponentHeight()) + (componentLibrary.getSize(node.getComponentType()).getHeight() / 2);
```
flo-dup: Now the shift you added depends on maxComponentHeight, that is if the maxComponentHeight is 100px you have the label 50px above the biggest component. You need to divide maxComponentHeight by 2 to have a "fixed" shift. And probably add an extra offset to avoid that the label is not touching the biggest component.

flo-dup (changes requested):

flo-dup (changes requested):
One thing which was missed out is that the current shift is applied on each node, not only on feeder nodes. It leads to an unwanted shift for labels on switches and internal pst for instance

flo-dup (approved):

SlimaneAmar (approved):

PR #761, opened 2025-12-04
Display feeder info on 3wt leg inside VL

### LINKED ISSUE #686: Display feeder info on the 3WT leg inside a VL

For a voltage level diagram, the feeder info values are displayed on both legs outside the voltage level and not on the leg inside the displayed voltage level.

For a voltage level diagram, the feeder info values are: 
- Else displayed on the leg inside the displayed voltage level only
- Or displayed on the three legs of the three-winding transformer

Both behaviors would be available through an option.

### Describe the motivation

It would fit the different needs of users (some want more information, some want less information / clearer diagrams)

### Extra Information

See also these issue and PR : 
- https://github.com/powsybl/powsybl-diagram/issues/565
- https://github.com/powsybl/powsybl-diagram/pull/567

quinarygio:

#686 

- [ ] Yes
- [ ] No

**Other information**:


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/CalculateCellHeightBlockVisitor.java
```diff
@@ -64,8 +66,10 @@ public void visit(BodyPrimaryBlock block) {
         // already encountered
         long nbNodes = block.getNodeStream().filter(n -> !encounteredNodes.contains(n) && n.getType() != BUS)
                 .count();
+        long nbSwitch3wt = block.getNodeStream().filter(n -> !encounteredNodes.contains(n) && n.getType() == NodeType.SWITCH)
+                .filter(sw -> sw.getAdjacentNodes().stream().anyMatch(n -> n instanceof Middle3WTNode)).count();
 
-        this.blockHeight = (nbNodes - 1) * componentHeight;
+        this.blockHeight = (2 * nbSwitch3wt + nbNodes  - 1) * componentHeight;
```
flo-dup: The calculation should take into account the space for arrows, use `LayoutParameters.getFeederSpan()` for that.
```suggestion
        this.blockHeight = (nbNodes  - 1) * componentHeight;
        if (blockWith3wt) this.blockHeight += layoutParameters.getFeederSpan();
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/CalculateCellHeightBlockVisitor.java
```diff
@@ -64,8 +66,10 @@ public void visit(BodyPrimaryBlock block) {
         // already encountered
         long nbNodes = block.getNodeStream().filter(n -> !encounteredNodes.contains(n) && n.getType() != BUS)
                 .count();
+        long nbSwitch3wt = block.getNodeStream().filter(n -> !encounteredNodes.contains(n) && n.getType() == NodeType.SWITCH)
```
flo-dup: I think you don't need to count these kind of switches, you only need to check if there's one 3wt in the block nodes. And if there is, you add one extra feeder span to the height.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/CalculateCoordBlockVisitor.java
```diff
@@ -53,20 +55,43 @@ public void visit(BodyPrimaryBlock block) {
...
+    private void setVerticalOrientationCoord(BodyPrimaryBlock block) {
+        List<Node> blockNodes = block.getNodes();
+        int sign = block.getOrientation() == UP ? 1 : -1;
+        double y0 = block.getCoord().get(Y) + sign * block.getCoord().getSpan(Y) / 2;
+        double yPxStep = sign * block.getCoord().getSpan(Y) / (blockNodes.size() - 1);
+        double swStep = sign * block.getCoord().getSpan(Y) / (blockNodes.size());
+
+        for (int i = 0; i < blockNodes.size(); i++) {
+            Node n = blockNodes.get(i);
+            if (n.getType() == NodeType.SWITCH && n.getAdjacentNodes().stream().anyMatch(Middle3WTNode.class::isInstance)) {
```
flo-dup: that's only working for some usecases. Indeed, it may be something else than a switch, and there might be more than one switch. If you try nominal_voltage_style_vl3.svg with two breakers before the disconnector, don't you end up with something strange?

I'm wondering if we could keep the code here unchanged but rather change the span of the block is there's one 3wt inside. Did you try something like that?

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/CalculateCoordBlockVisitor.java
quinarygio: I already changed the span of the block when there is a 3wt inside. Now I changed the condition to change the position of all the switches and feeders and added a test with more than one switch.

single-line-diagram/single-line-diagram-core/src/test/resources/Test3WTMultiSwitchFeederInfoArrowVoltageLevel.svg
flo-dup, line 1: This unit test reference highlights the calculation problem: the switches are not layout at the expected positions. See the below screenshot with the show internal nodes SvgParameter enabled:

Image

NathanDissoubray: image

should the internal node be lower, where I drew a red circle ? or should the switch closest to the busbar be higher ? Also shouldn't there be an internal node between the two switches, where I pointed with the arrow ?

PR #774, opened 2026-01-20
Introduce an ID provider for SVGs

### LINKED ISSUE #772: add mRIDs in NAD to enable interactivity

Continuation of
- #754 

The Talk2PowerSystem chatbot can display powsybl diagrams, and next we'll add interactivity: when the user clicks on an electrical element, the bot will state some info about it, and ask what else the user wants to know about it
- https://github.com/statnett/Talk2PowerSystem/blob/main/diagram/svg-click#readme is an analysis of SVG `id` appearing in SLDs
- So now I know which SVG elements correspond to a mRID (and now I see that #754 is a bit naive). 
  - https://github.com/statnett/Talk2PowerSystem/tree/main/diagram/svg-click#sld-diagrams-clickable shows that (the red-on-black diagram)
  - See https://github.com/statnett/Talk2PowerSystem/blob/main/diagram/svg-click#summary for my decisions (feedback is appreciated!)
- However, https://github.com/statnett/Talk2PowerSystem/tree/main/diagram/svg-click#nad-diagrams-not-clickable shows that NAD diagrams are not clickable at all:
  - their SVG `id` are sequential and don't include any mRID.

Would it be possible to improve this? Many of the NAD SVG elements correspond to a single electrical element...

### Describe the motivation

_No response_

### Extra Information

_No response_

rolnico:

No

Feature

In SLD, the ids used in the SVGs are created by prepending the network element ids with `id` and by replacing non-alphanumeric characters with their unicode value.

An `IdProvider` can now be given to the `draw` methods to specify how the ids should be generated.
Default behavior is still the same, using `LegacyIdProvider`. A new IdProvider is also available, it works as in NAD and generates the ids by counting the elements.

- [ ] Yes
- [x] No

**Other information**:


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -473,16 +462,14 @@ protected void setMetadata(GraphMetadata metadata, Node node, String nodeEscaped
             }
         }
 
-        String prefixId = metadata.getSvgParameters().getPrefixId();
-
         String vId = graph instanceof VoltageLevelGraph voltageLevelGraph ? voltageLevelGraph.getVoltageLevelInfos().id() : "";
 
-        boolean isOpen = node.getType() == NodeType.SWITCH && ((SwitchNode) node).isOpen();
+        boolean isOpen = node instanceof SwitchNode switchNode && switchNode.isOpen();
```
alicecaron, line 467: Not directly related to this development but there are other places where this test is made and could be replaced:
`node.getType() == NodeType.SWITCH` => `node instanceof SwitchNode`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
```diff
@@ -1108,7 +1062,7 @@ protected void createDefsSVGComponents(Document document, Set<String> listUsedCo
                 Map<String, List<Element>> subComponents = componentLibrary.getSvgElements(c);
                 if (subComponents != null && !subComponents.isEmpty()) {
                     Element group = document.createElement(GROUP);
-                    group.setAttribute("id", c);
+                    group.setAttribute("id", c); // TODO: check here
```
alicecaron: todo to solve? what needs to be checked?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/SingleLineDiagram.java
```diff
@@ -102,52 +102,100 @@ public static void drawVoltageLevel(Network network, String voltageLevelId, Path
         drawVoltageLevel(network, voltageLevelId, svgFile, new SldParameters());
     }
 
+    public static void drawVoltageLevel(Network network, String voltageLevelId, Path svgFile, IdProvider idProvider) {
```
alicecaron, line 122: method is not tested

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultSVGWriter.java
rolnico: Indeed, we can. I opened #805 for this.

alicecaron (dismissed):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultIdProvider.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public String getOrCreateSvgId(String equipmentId) {
+        return getOrCreateId("equipmentId_" + equipmentId);
+    }
+
+    @Override
+    public String getOrCreateSvgId(String equipmentId, String subType) {
+        return getOrCreateId(String.format("equipmentId_%s_subType_%s", equipmentId, subType));
```
flo-dup, line 34: Putting information into a String is bad design... And it could lead to conflicts. For instance, in this case you've got the problem of an equipmentId "ALDJF_subtype_AER" (leading to entry "equipmentId_ALDJF_subtype_AER") clashing with another equipment id "ALDJF" with subtype "AER" (leading again to entry "equipmentId_ALDJF_subtype_AER").

The problem already lies within the `getOrCreateSvgId` goal. When don't we know if it's a create or a get when we call it? Maybe the idProvider could be responsible for having the map allowing to get the id from any object, but isn't this information already in the metadata objects?

flo-dup (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultIdProvider.java
NathanDissoubray: Should the logic of the map be put in an `AbstractIdProvider` so that `LegacyIdProvider` and `DefaultIdProvider` extend from that ?
I'm not sure to understand what you mean by "When don't we know if it's a create or a get when we call it?"
How do you propose we ensure the unicity of the naming, because as it stands with the string we can't guarantee it.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/LegacyIdProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.powsybl.sld.util.IdUtil;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Florian Dupuy {@literal <florian.dupuy at rte-france.com>}
+ */
+public class LegacyIdProvider implements IdProvider {
```
NathanDissoubray, line 18: Would `CountingIdProvider` be better for the class name ? I think it's better to give information about the function of this class rather than the fact it's old.

single-line-diagram/single-line-diagram-core/src/test/resources/TestDefaultIdProvider.svg
```diff
@@ -0,0 +1,534 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
NathanDissoubray, line 1: we can see that the text of the ids of B11_B12_1 and B11_B12_2 overlap with the line (or rather is hidden by it). I don't think it's an issue with this PR directly, but shouldn't the positioning of the label of the id account for the horizontal span of said ID ?

single-line-diagram/single-line-diagram-core/src/test/resources/TestDrawS1PrefixLegacy.svg
```diff
@@ -0,0 +1,676 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
NathanDissoubray, line 1: appart from all the `idLegacy` becoming `idTest` (which is a difference in the text format of the svg), there is no visual difference between `TestDrawS1PrefixLegacy.svg` and `TestDrawS1PrefixTest.svg`. What is the goal of having those 2 files ?

NathanDissoubray (commented):
There are some files in the SVG that look very similar and just have different svg class id (so no visual difference). I'm wondering what's the goal with those.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultIdProvider.java
flo-dup: To ensure the unicity of the naming: just use a counter++ like in network-area-diagram, that's enough. And for the legacy id provider, we don't need to do anything as we cannot ensure anything without breaking it.

To me there should be no map. The IdProvider should be called only for creating id, not for getting an already created id. The nodes and edges should rather hold their id, hence no map is needed. The nodes already hold an internal id in fact, which is currently useless, so we could make good use of it. I think a good design would be to use the IdProvider at nodes and edges creation.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/LegacyIdProvider.java
flo-dup: It's not based on counting.

I named it "legacy" as it gives the information that it replicates the previous behaviour and as it implies that you should rather not use it.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultIdProvider.java
flo-dup: (that was already discussed with @rolnico and we agreed that the underlying changes were too significant to finish this PR juste before the release)

single-line-diagram/single-line-diagram-core/src/test/resources/TestDefaultIdProvider.svg
flo-dup: we do not know the horizontal span of a text in a svg produced by single-line-diagram, as the users specify the font size with an external CSS

single-line-diagram/single-line-diagram-core/src/test/resources/TestDrawS1PrefixLegacy.svg
flo-dup: We probably need to change all the unit test references to use the default counting id provider once this PR is finished (maybe in a separate PR to have it reviewed more easily), hence the two files

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/LegacyIdProvider.java
NathanDissoubray: Got confused with the default which has a count

Issue #762, opened 2025-12-05
[NAD] Enrich the metadata to allow creating a diagram’s SVG from the metadata alone


CBiasuzzi:
### Describe the current behavior

Currently, powsybl-diagram allows us to create a NAD SVG and, optionally, a JSON metadata. The metadata, in addition to the SVG, is for example used by powsybl-network-viewer  to support its advanced display and interactive features.

### Describe the expected behavior

We’d like to have the option to delegate the creation of the SVG to the viewer component, using just the metadata (e.g. for performance improvement, especially for large networks).

However, info about the NAD in the current metadata is not sufficient to create the SVG from scratch.  For example: some invisible nodes in partial networks are filtered out by powsybl-diagram and not included in the SVG, while the metadata includes also the invisible ones it’s impossible to tell which node is visible and which is not.

Therefore, we’d like to enrich the NAD metadata schema with more details, including for instance:

Node characterization:
- Hidden node
- Node having an unknown bus
- 3WT node
- Boundary node
 
CSS classes assigned to the NAD elements (nodes, buses, edges)

flo-dup:
> Another information that could be useful is if and edge is disconnected. This information is used, at least, for adding a specific CSS class nad-disconnected to the edge element

As we need to add a "style" or "classes" field to the edge and node metadata, the nad-disconnected class would already be included there. But you say "at least", this might be reconsidered if there's more than style for disconnected lines - do you have in mind some other use of the disconnected information?

massimo-ferraro:
In a network with voltage levels with unknown bus, in the metadata there is no bus metadata for these unknown buses 
But in the SVG, in the circle elements representing the VLs, there is a CSS class (besides the nad-unknown-busnode class):
```
<g transform="translate(-312.98,-477.87)" id="8" class="nad-vl300to500">
    <circle class="nad-unknown-busnode nad-bus-0" r="40.00"/>
</g>
```
This class nad-bus-0 is not in the metadata, so far, also in the enriched metadata (at least non related to this VL).
Is there any other way the viewer component in charge of creating the SVG could get this information? 
If added to the diagram metadata, where should this CSS class be put? in the VL node metadata? Where else?

CBiasuzzi:
this is a good point! It's actually true that, by design, the unknown bus (BusNode.UNKNOWN) is not serialized in the metadata bus list.
[...]

PR #781, opened 2026-02-05
[NAD] enrich metadata

CBiasuzzi:

Implements #762

- [ ] Yes
- [ ] No

This PR includes some refactoring to decouple style computation from rendering.  SvgWriter and DiagramMetadata now read styles and class attributes directly from Graph elements, rather than invoking a style provider directly.  Consequently, the style provider must be applied to the graph before the rendering phase.
Also, the style-application must occur after the layout phase: certain attributes (e.g., bus index for TopologicalStyleProvider) are only available post-layout, so applying the style later ensures they are correctly set.

Code calling  SvgWriter or DiagramMetadata directly after graph construction must now apply layout and styles (e.g., using `NetworkGraphBuilder.applyStyle(graph, styleProvider)`) beforehand .

**Other information**:


network-area-diagram/src/main/java/com/powsybl/nad/svg/metadata/NodeMetadata.java
```diff
@@ -38,7 +41,10 @@ public NodeMetadata(@JsonProperty("svgId") String svgId,
                         @JsonProperty("legendSvgId") String legendSvgId,
                         @JsonProperty("legendEdgeSvgId") String legendEdgeSvgId,
                         @JsonProperty("legendHeader") List<String> legendHeader,
-                        @JsonProperty("legendFooter") List<String> legendFooter) {
+                        @JsonProperty("legendFooter") List<String> legendFooter,
+                        @JsonProperty("boundary") boolean boundary,
+                        @JsonProperty("threeWindings") boolean threeWindings,
+                        @JsonProperty("unknownBus") boolean unknownBus) {
```
flo-dup: Couldn't it be a type enum instead of 3 booleans? Or at least for the two first ones?

CBiasuzzi: Thank you for the suggestion:  it makes sense, since boundary and threeWindings are mutually exclusive.
I just wonder if the name 'type' would be the best choice for  the attribute; 'type' seems to be quite 'committing' ... for an optional attribute.
Unless we make it not optional. What if we make the  'type' attribute in NodeMetadata  always available and filled with simply the getClass().getSimpleName()  for the Node instance.
What do you think?

flo-dup: Sorry, missed your comment! We could make it non-optional but indicate a default value to Jackson to have a serialization which is not too verbose?

CBiasuzzi: It's true that in the vast majority of cases the  type would  be VoltageLevelNode, so to not ... bloat ... the metadata file I wonder if it would be actually better to keep this 'type' optional, as in your original suggestion

network-area-diagram/src/main/java/com/powsybl/nad/NetworkAreaDiagram.java
```diff
@@ -84,12 +86,13 @@ public static void draw(Network network, Writer writer, Writer metadataWriter, N
         Objects.requireNonNull(param);
 
         Graph graph = getLayoutResult(network, param, voltageLevelFilter);
-        createSvgWriter(network, param).writeSvg(graph, writer);
-        createMetadata(graph, param).writeJson(metadataWriter);
+        StyleProvider styleProvider = param.getStyleProviderFactory().create(network);
+        createSvgWriter(param, styleProvider).writeSvg(graph, writer);
+        createMetadata(graph, param, styleProvider).writeJson(metadataWriter);
```
flo-dup: This approach results in twice the calls to styleProvider. It remains valid of course but it is different from the approach chosen for label provider when labels were included in metadata (#740): labelProvider is given to networkGraphBuilder, the graph then creates the label information on each edge and on each node, and finally the SvgWriter includes it in the SVG based solely on the graph.

I think we should align both approaches, don't you think?

The cleanest way would probably be to first build DiagramMetadata, then call the SvgWriter based on it. This would mimic the network-viewer which creates the SVG based on metadata. But that's a third way to do it, which probably changes the logic too much for current PR. If we plan to do so, we could keep PR current state instead of aligning with the previous labelFactory approach.

CBiasuzzi: I updated the PR as suggested (first suggestion): now the style classes are first added to the graph, then used by both SvgWriter and DIagramMetadata. I added a note to the PR description.

network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java
```diff
@@ -113,13 +111,18 @@ private void writeSvg(Graph graph, OutputStream svgOs) {
         Objects.requireNonNull(graph);
         Objects.requireNonNull(svgOs);
 
+        // Style expected to be already applied to the graph
+        if (!graph.isStyleApplied()) {
+            throw new PowsyblException("Style must be applied to the graph");
+        }
```
rolnico, line 117: What if, in order to keep the compatibility, we do:

```suggestion
        if (!graph.isStyleApplied()) {
            NetworkGraphBuilder.applyStyle(graph, styleProvider);
        }
```
and keep the styleProvider as parameter to the constructor?

Or maybe have one constructor with the styleProvider and one without (then defining `styleProvider=null`) and then having here:

```suggestion
        if (!graph.isStyleApplied()) {
            if (styleProvider != null) {
                NetworkGraphBuilder.applyStyle(graph, styleProvider);
            } else {
                throw new PowsyblException("Style must be applied to the graph");
            }
        }
```

@flo-dup what do you think of this?

rolnico: At the start of the class, we would then have:

```java
    private final SvgParameters svgParameters;
    private final StyleProvider styleProvider;
    private final EdgeRouting edgeRouting;
    private final NadComponentLibrary componentLibrary;

    public SvgWriter(SvgParameters svgParameters, NadComponentLibrary componentLibrary,
                    EdgeRouting edgeRouting) {
        this.svgParameters = Objects.requireNonNull(svgParameters);
        this.styleProvider = null;
        this.componentLibrary = Objects.requireNonNull(componentLibrary);
        this.edgeRouting = Objects.requireNonNull(edgeRouting);
    }

    public SvgWriter(SvgParameters svgParameters, StyleProvider styleProvider,
                     NadComponentLibrary componentLibrary, EdgeRouting edgeRouting) {
        this.svgParameters = Objects.requireNonNull(svgParameters);
        this.styleProvider = styleProvider;
        this.componentLibrary = Objects.requireNonNull(componentLibrary);
        this.edgeRouting = Objects.requireNonNull(edgeRouting);
    }
```

rolnico: Seen with @flo-dup : no need to keep the compatibility

flo-dup: I think there's no compatibility need, so we could keep it like it is in current state

PR #807, opened 2026-04-21
Fix division by 0 issue in Force layout repulsion calculation

NathanDissoubray:

None

Fix a division by 0 leading to NaN dimensions of the force layout bounding box (and NaN position of some points)

Division by 0 can occur.

Check if the magnitude is 0 before dividing.
Adding a vector of `(1, 1)` instead of an arbitrarily high value, so the actual force doesn't send the two points flying away at high speed (if we use `(1000, 1000)` for example, it might be way too much for a small graph). The actual repulsion will occur on the next time step (with realistic force values), as long as the point are separated.
Unless this happens on the last time step of the post-processing, but that would be really unlucky.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinear.java
+++ b/diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinear.java
@@ -13,7 +13,6 @@ import com.powsybl.diagram.util.layout.geometry.Point;
 import com.powsybl.diagram.util.layout.geometry.Vector2D;
 
 import java.util.Map;
-import java.util.Random;
 
 /**
  * A repulsion force between a point and all the other points of the graph. It is simpler to calculate than the coulomb force.
@@ -24,7 +23,6 @@ public class RepulsionForceDegreeBasedLinear<V, E> extends AbstractDegreeBasedFo
 
     private final double forceIntensity;
     private final boolean effectFromFixedNodes;
-    private Random random;
 
     public RepulsionForceDegreeBasedLinear(double forceIntensity, boolean effectFromFixedNodes) {
         this.forceIntensity = forceIntensity;
--- a/diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinear.java
+++ b/diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinear.java
@@ -63,18 +63,24 @@ public class RepulsionForceDegreeBasedLinear<V, E> extends AbstractDegreeBasedFo
     ) {
         // The force goes from the otherPoint to the point (repulsion)
         Vector2D force = Vector2D.calculateVectorBetweenPoints(otherPoint, point);
-        // divide by magnitude^2 because the force multiplies the unit vector by something/magnitude
-        // the unit vector is Vector/magnitude, thus the force is Vector/magnitude * something/magnitude, thus Vector/magnitude^2
-        // if we just use the vector and not the unit vector, points that are further away will have the same influence as points that are close
-        // this is easy to explain as the formula is Vector * k * deg(n1) * deg(n2)/distance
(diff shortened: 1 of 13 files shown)


diagram-util/src/main/java/com/powsybl/diagram/util/layout/forces/RepulsionForceDegreeBasedLinear.java
```diff
@@ -63,18 +65,27 @@ private void linearRepulsionBetweenPoints(
...
                 * (otherPoint.getPointVertexDegree() + 1)
-                / force.magnitudeSquare();
-        force.multiplyBy(intensity);
-        resultingForce.add(force);
+                / magnitudeSquare;
+            force.multiplyBy(intensity);
+            resultingForce.add(force);
+        } else {
+            if (random == null) {
+                random = new Random(45L);
```
samirromdhani: SonarCloud Code Analysis fail, 3 security hotspot for the same thing, you can avoid duplicate code.

NathanDissoubray: it fails because it thinks there is a problem with using non cryptographically secure random (which we don't really care here)
I'll see what I can do for duplicate code, but since it's very small, it's possible removing the duplication will actually need more code

NathanDissoubray: I tried making an util class to get the random force value from, but since it's static and called by everything, depending on the order of the tests the results can change (because if test A calls the random before test B, or vice-versa, they get different values). This problem happens specifically because it's static.

If I don't make it static, I need to creates instances of the `RandomForce` in each force class when needed (and thus we get back to checking if it's null or not). What we gain by doing this is not having the line `new Vector2D(random.nextDouble(1, 2), random.nextDouble(1, 2))` repeated in three different files, but we also need an entire class.
Do you have any idea to make this work ? Otherwise, I think having a few lines repeated across 3 files is preferable

samirromdhani: Here, the gain is about making the random initialization centralized (new Random() in one place),
We can also keep the code extensible with `RandomGenerator` interface, and by default use Random implementation

Some others benefits, 
- In tests we can provide other implementation if needed like `SplittableRandom`
- Algrithms depond only on interface, and the concrete random generator is configured in a single place (in RandomForce)

Here is an example, https://github.com/powsybl/powsybl-diagram/commit/929e4efe61b2454289c211e73de5c3fbadd656e7

samirromdhani: Branch : https://github.com/powsybl/powsybl-diagram/tree/test/807_proposal_random_generator

NathanDissoubray: I took a look at your branch, thank you for the idea.
This would mean this PR is a breaking change since we are changing the constructor of some forces.
What would be the goal of being able to give another random ? What benefit would another random implementation give us for the goal we are trying to achieve ?

rolnico: I propose the following changes
[Move_Random_to_LayoutContext_to_have_deterministic_results.patch](https://github.com/user-attachments/files/27432127/Move_Random_to_LayoutContext_to_have_deterministic_results.patch)

Moving the Random to the LayoutContext allows to have reproductible and consistent results, while avoiding breaking changes.

NathanDissoubray: that seems like the best idea thus far, let's go with that :+1: used your patch in 3b3a35d2a0fc

rolnico (approved):

PR #811, opened 2026-05-11
Add information to the SLD and NAD metadata

### LINKED ISSUE #801: [NAD] Add network name in the metadata

The generated diagram (SVG + JSON) does not contain any information about the network displayed (name, date)

The generated diagram (SVG + JSON) contains information about the network name, at least the name

### Describe the motivation

When exporting the diagram in png or svg with powsybl-network-viewer, the resulting file cannot be named better than nad.svg / nad.png.

### Extra Information

_No response_

### LINKED ISSUE #802: Add powsybl-diagram current version number in the metadata

There's no information about the version used to generate a given SVG, even with its JSON metadata file

Get the powsybl-diagram version used to generate a diagram within the metadata file.

### Describe the motivation

Use that version number for checking the compatibility in powsybl-network-viewer

### Extra Information

_No response_

rolnico:

Fixes #801 
Fixes #802

Feature

- [ ] Yes
- [x] No


diagram-util/src/main/java/com/powsybl/diagram/metadata/AbstractMetadata.java
```diff
@@ -14,26 +14,60 @@
...
+                return v.getMavenProjectVersion();
+            }
+        }
+        return null;
+    }
+
+    @JsonProperty("diagramVersion")
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public String getDiagramVersion() {
+        return diagramVersion;
```
flo-dup: I think it's enough to simply have that method added with following content:

```suggestion
        return new PowsyblDiagramVersion().getMavenProjectVersion();
```

As we're not supporting writing older versions of SVG, giving access to a method which overrides diagram version is not appropriate. I get that for unit tests it's easier to do so, but I think we can avoid it (mocking the version number? adding it in the reference when comparing? ...)

flo-dup (changes requested):

network-area-diagram/src/main/java/com/powsybl/nad/svg/metadata/DiagramMetadata.java
```diff
@@ -52,9 +56,17 @@ public enum NodeType {
     @JsonSetter(nulls = Nulls.AS_EMPTY) // if missing when deserializing creates an empty array
     private final List<InjectionMetadata> injectionsMetadata = new ArrayList<>();
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private String networkId;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private String networkName;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private String networkDate;
+
```
NathanDissoubray: should this be in `AbstractMetadata` rather than here, so it can also be used in the SLD and not just in the NAD ?

diagram-util/src/main/java/com/powsybl/diagram/metadata/AbstractMetadata.java
```diff
@@ -15,13 +15,70 @@
...
+    public String getNetworkId() {
+        return networkId;
+    }
+
+    @JsonProperty("networkDate")
+    public String getNetworkDate() {
+        return networkDate;
+    }
+
+    protected T setNetworkInformation(Network network) {
```
flo-dup: The method is convenient but it couples AbstractMetadata to IIDM even though the class only needs three simple values. Besides, the metadata class should rather describe serialized metadata, not know how to extract it from a domain object.

So-Fras (dismissed):
Thank you for the work! 
I think that this PR shows that a refactoring is welcome on the API side (both in SLD and NAD).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
```diff
@@ -405,4 +416,9 @@ public SvgParameters getSvgParameters() {
         return svgParameters;
 
     }
+
+    @Override
+    public GraphMetadata setNetworkInformation(Network network) {
```
flo-dup: No need to override. The method in AbstractMetadata should be public though.

network-area-diagram/src/main/java/com/powsybl/nad/svg/metadata/DiagramMetadata.java
```diff
@@ -109,6 +120,11 @@ public SvgParameters getSvgParameters() {
         return svgParameters;
     }
 
+    @Override
+    public DiagramMetadata setNetworkInformation(Network network) {
```
flo-dup: Same thing, no need to override

flo-dup (dismissed):

diagram-util/src/main/java/com/powsybl/diagram/metadata/AbstractMetadata.java
NathanDissoubray: Should I remove it from `AbstractMetadata` then ? Or change it to require the three strings for the network name, id and date instead of the network ?

flo-dup: Change it to require the three strings (or a small NetworkInfo record containing them if you want)

NathanDissoubray: I'll still need the network at one point in the SLD/NAD to get those informations
And if I use 3 strings, I'll actually do the same that is written in this function (but twice, one in NAD and one in SLD).
Or I could add 3 parameters to some functions of the SLD and NAD, but those functions already have enough parameters I think.

NathanDissoubray: Done in c36058399678

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/GraphMetadata.java
NathanDissoubray: Done in c36058399678

network-area-diagram/src/main/java/com/powsybl/nad/svg/metadata/DiagramMetadata.java
NathanDissoubray: Done in c36058399678

So-Fras (approved):

PR #813, opened 2026-05-18
[SLD] Add new separate option to write bus labels diagonally

rolnico:

No

Feature/Bug

Users can either have all labels diagonal or none.

Users can choose to have bus labels and/or feeder labels diagonal.

- [ ] Yes
- [x] No

**Other information**:


sBouzols (approved):
Code Review OK
Tests OK in GridStuite project

flo-dup:
The labelDiagonal parameter is not documented, if I try to write some that would something like: "Enable the labels to be displayed rotated with an angle provided in angleLabelShift parameter. But this does not apply to bus labels, which are rotated only iff busLabelDiagonal is set to true. And angleLabelShift != 0. Note that the angle between labels is the same. Note that for labels other than buses and feeders, the behaviour is not guaranteed" Such a complex documentation shows there's at least a naming problem in those parameters.

Issue #814, opened 2026-05-19
Improve snakeline path in multi-substation SLD


NathanDissoubray:
### Describe the current behavior

Snakeline pathing in multi-substation SLD tends to overlap lines between the substation and use uneeded turns, which makes the image difficult to read. Example:

<img width="1590" height="472" alt="Image" src="https://github.com/user-attachments/assets/28ece79d-6829-4c59-8cfb-f06c80135a6c" />

In red the overlaping / crossing lines, in blue the turns that are not needed.

### Describe the expected behavior

Lines should not overlap or cross when it is clear that it shouldn't, Avoiding crossing in the general case is not always possible, but in simple cases, this shouldn't be an issue. The same network as above, with an A* path search (originally made in #730)

<img width="1612" height="490" alt="Image" src="https://github.com/user-attachments/assets/389096f0-486e-4614-a389-3692322aafaa" />

We can see there is much less overlapping / crossing, and no uneeded turns.

### Describe the motivation

Provide a better visual for multi-substation SLD to make it more readable.

### Extra Information

An attempt has been made in #730 
Although the visual is much better, it also takes much more time to calculate the path between the substation (40-45s to run all SLD test, against 5-8s currently). The current code uses a Dijkstra algorithm, but it does it faster by creating lines on the canvas that are the only path allowed for lines between substations. This technique has the negative effect of creating more overlapping and uneeded turns. It also has a termination bug where it will use the first path it finds, regardless of the score of that path (even if a path of better score can be found).

#730 does not use this kind of hack, and allows the entire canvas to be searched for paths by an A*. It leads to better visuals, but it also takes longer to calculate.

An improvement can be made on A*. There exists multiple algorithms in the litterature that do this kind of search, such as Jump Point Search (JPS). The issue with most of those algorithms is that they consider path as turn symmetric, meaning that two path of some length have the same cost, regardless of the number of turns of each path. For our use case, we want to minimize the number of turns, so this assumption is not verified.

An adaptation of Rectangle Expansion A* (Zhang A et al. Rectangle expansion A* pathfinding for grid maps, Chin J Aeronaut (2016)) that is non turn-symmetric can be made. It would be much faster as most of the shapes in the SLD are rectangular or close to rectangular, which is ideal for the rectangle expansion, as it can cover large parts of the canvas quickly. This is the implementation I would go with, and I have started to implement it.

FredericSabot:
[...]

PR #817, opened 2026-06-02
[SLD] Allow manual positioning of substation in multi substations diagrams

### LINKED ISSUE #809: [SLD] Allow manual positioning of substation in multi substations diagrams

Multi substations diagrams only allow for a matrix layout

Allow manually set the position of (one or multiple of) the substations on the diagram

### Describe the motivation

This would be useful in the TwinEU project (allow operators to customize their network view).

### Extra Information

I will try to work on this next week.

Ideally, this should also be integrated in the powsybl-network-viewer.

FredericSabot:

Closes #809 

- [x] Yes
- [ ] No

The public method computePathFindingGrid in MatrixZoneLayout has been removed. An equivalent method is now (privately) automatically called by the base class AbstractPositionedZoneLayout (parent of MatrixZoneLayout and the new ManuallyPositionedZoneLayout). This should have no impact on users.

**Other information**:

Can be tested with

```java
Network network = FourSubstationsNodeBreakerFactory.create();

List<Pair<String, Point>> positions = List.of(
    Pair.of("S1", new Point(100, 100)),
    Pair.of("S2", new Point(600, 100)),
    Pair.of("S3", new Point(300, 400))
);

SldParameters sldParameters = new SldParameters()
    .setZoneLayoutFactory(new PositionedZoneLayoutFactory(positions));

SingleLineDiagram.drawMultiSubstations(network, positions.stream().map(Pair::getFirst).toList(), Path.of("out-positioned.svg"), sldParameters);
```

Which generates

out-positioned

The calculatePolylineSnakeLine function has been loosely copied from MatrixZoneLayout, however, it would be best to update the pathfinding to not depend on the layout type (in another PR).


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,308 @@
...
+
+        // Horizontal strip across the substation + margin so paths can route laterally
+        int ssX = (int) getSubstationOrigin(sg).getX();
+        int ssW = (int) sg.getWidth();
+        double exitY = d == Direction.TOP ? minY : maxY;
+        for (int x = ssX - margin; x < ssX + ssW + margin; x++) {
+            pathFinderGrid.setAvailability(x, exitY, true);
+        }
+    }
+
```
NathanDissoubray, line 248: it might be better to change those method as protected in `MatrixZoneLayoutModel` and then make `PositionedZoneLayout` extend from it.
Another possibility would be to create an abstract class, moving useful methods from `MatrixZoneLayoutModel` to this abstract class, and then making both `MatrixZoneLayoutModel` and `PositionedZoneLayout` extend from the abstract class.
This abstract class could be `AbstractZoneLayout` or a class that extends from it, if `AbstractZoneLayout` is still too general

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,308 @@
...
+    private boolean overlaps(double[] a, double[] b, int margin) {
+        boolean xOverlap = a[0] < b[0] + b[2] + margin && a[0] + a[2] + margin > b[0];
+        boolean yOverlap = a[1] < b[1] + b[3] + margin && a[1] + a[3] + margin > b[1];
+        return xOverlap && yOverlap;
+    }
+
+    private void nudge(double[] a, double[] b, int margin) {
```
NathanDissoubray: Could you change the name of this function so it's clearer what it does ? If you can also add some javadoc to it, it would be clearer I think

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,308 @@
...
+                    if (overlaps(a, b, margin)) {
+                        nudge(a, b, margin);
+                        moved = true;
+                    }
+                }
+            } while (moved);
+        }
+    }
+
+    private boolean overlaps(double[] a, double[] b, int margin) {
```
NathanDissoubray: same here regarding naming and javadoc

NathanDissoubray (changes requested):
I think some code between `MatrixZoneLayoutModel` and `PositionedZoneLayout` could be moved in a common class, which would prevent this code to be "loosely copied from MatrixZoneLayout" as you put it yourself.
This will make it overall cleaner, easier to maintain and to review.

You should also add tests. Think of using `mvn clean install` to have checkstyle check your code, you have a dozen double blank spaces problems.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: I was not too sure how long this code would live given the pathfinding updates, but ok, will update this.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: Fixed

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: Fixed

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: The MatrixZoneLayout code was a bit different since it used matrix instead of positions, so I have merged both into an AbstractStrictlyPositionedClass whose the substation positions are computed by the children. This causes some tests to fail with small diffs on the polylines and diagram size. I will let you decide whether this is acceptable.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
```diff
@@ -0,0 +1,234 @@
...
+
+import java.util.*;
+
+/**
+ * Zone layout that places substations at desired top-left coordinates (x, y).
+ * <p>
+ * Each substation is assigned a desired top-left position via a {@code List<Pair<String, Point>>}.
+ * After running the sub-layouts (which determine each substation's pixel size),
+ * substations that would overlap — including the snakeline margin — are nudged
+ * in list order: earlier entries win and later ones are moved away.
```
NathanDissoubray: could you reformulate this to explain the term "nudge" ? do you mean to say that the substations are moved to their respective positions ? what do you mean by "win" ? position the substations in the list order, and if two substations overlap, then the one with the higher index is moved ?
Also, why would there be overlap if we calculate substation position, can't we calculate the positions to prevent the overlap in the first place ?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: I have committed the diffs for reference, but they can be reverted.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @author Frédéric Sabot
+ */
+public abstract class AbstractStrictlyPositionedZoneLayout extends AbstractZoneLayout {
```
NathanDissoubray: I think the name could be `AbstractPositionedZoneLayout`. What is the goal of the `Strictly` here ?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Zone layout that places substations at desired top-left coordinates (x, y).
+ * <p>
+ * Each substation is assigned a desired top-left position via a {@code List<Pair<String, Point>>}.
+ * After running the sub-layouts (which determine each substation's pixel size),
+ * substations that would overlap — including the snakeline margin — are nudged
+ * in list order: earlier entries win and later ones are moved away.
+ *
+ * @author Frédéric Sabot
```
NathanDissoubray: could you include your e-mail in this format : 
```
@author Thomas Adam {@literal <tadam at neverhack.com>}
```
example taken from `MatrixZoneLayout`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,234 @@
...
+                        Grid.getPointsAlongSnakeline(branch.getSnakeLine())
+                            .forEach(p -> pathFinderGrid.setAvailability(p, false));
+                    }
+                });
+            });
+            sg.getLineEdges().forEach(e ->
+                Grid.getPointsAlongSnakeline(e.getSnakeLine())
+                    .forEach(p -> pathFinderGrid.setAvailability(p, false)));
+        }
+    }
```
NathanDissoubray, line 218: I notice that some of the code is different from `MatrixZoneLayoutModel`. Is there a reason for not copying directly the code that was removed from `MatrixZoneLayoutModel` ? I think this is why test result are different, and I don't think they should be different if possible.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,157 @@
...
+
+import java.util.*;
+
+/**
+ * Zone layout that places substations at desired top-left coordinates (x, y).
+ * <p>
+ * Each substation is assigned a desired top-left position via a {@code List<Pair<String, Point>>}.
+ * After running the sub-layouts (which determine each substation's pixel size),
+ * substations that would overlap — including the snakeline margin — are nudged
+ * in list order: earlier entries win and later ones are moved away.
```
NathanDissoubray: I think "nudged" is not clear enough. This explanation is better than the one in `AbstractStrictlyPositionedZoneLayout` and explains better how it works.
How are the substation moved ? Left / Right ? Minimal distance from originally desired position ? I think it would be good to explain it here.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,157 @@
...
+ * @author Frédéric Sabot
+ */
+public class PositionedZoneLayout extends AbstractStrictlyPositionedZoneLayout {
+
+    /** Mutable bounding box used during overlap resolution. */
+    private static class Rectangle {
+        double x;
+        double y;
+        final double w;
+        final double h;
```
NathanDissoubray: use width, height instead of w, h

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,157 @@
...
+    /** Desired top-left (x, y) per substation, in priority order for overlap resolution. */
+    private final List<Pair<String, Point>> desiredPositions;
+
+    protected PositionedZoneLayout(ZoneGraph graph,
+                                   List<Pair<String, Point>> desiredPositions,
+                                   ZoneLayoutPathFinderFactory pathFinderFactory,
+                                   SubstationLayoutFactory sLayoutFactory,
+                                   VoltageLevelLayoutFactory vLayoutFactory) {
+        super(graph, pathFinderFactory, sLayoutFactory, vLayoutFactory);
+        this.desiredPositions = List.copyOf(Objects.requireNonNull(desiredPositions));
```
NathanDissoubray: why do you need to copy the list ? i don't see any code that modifies the original list here.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,157 @@
...
+     * Repeat until no overlaps remain with any earlier substation.
+     */
+    private void resolveOverlaps(List<Rectangle> rectangles, int margin) {
+        for (int i = 1; i < rectangles.size(); i++) {
+            boolean moved;
+            do {
+                moved = false;
+                for (int j = 0; j < i; j++) {
+                    Rectangle a = rectangles.get(j); // winner (fixed)
+                    Rectangle b = rectangles.get(i); // candidate (may move)
```
NathanDissoubray: could you change the name of the rectangles so that the comment is not needed to understand the code ?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,157 @@
...
+                    if (a.overlaps(b, margin)) {
+                        moveToResolveOverlap(a, b, margin);
```
NathanDissoubray: since you solve the overlap 1 by 1, I can imagine the following situation where we would never solve the overlap:

Image

To solve this, you can either:
- force the direction of the move (always move right or left, depending on which one forces the least overlap), I'm not sure this would solve the issue
- always move the external substations (the ones that are most, left, right, top or down), first. This would have for effect to "scale" the space between the substations
- there are algorithms that solve this kind of problem, since this is a well known problem: https://en.wikipedia.org/wiki/Rectangle_packing#Packing_different_rectangles_in_a_minimum-area_rectangle. You could try implementing one of those algorithms

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
```diff
@@ -186,4 +189,26 @@ void testInvalidZoneGraphMatrix() {
...
     }
+
+    @Test
+    void testZoneGraphPositioned() {
+        // build zone graph
+        network = Networks.createNetworkWithManySubstations();
+        List<Pair<String, Point>> positions = List.of(
+            Pair.of("A", new Point(100, 100)),
+            Pair.of("B", new Point(800, 100)),
+            Pair.of("C", new Point(300, 600))
```
NathanDissoubray: does this case test the overlap resolution ?

NathanDissoubray (changes requested):
To answer your question, I don't think that adding a new layout should change the results of the existing layouts. As such, I would prefer if you could find a solution that did not modify the existing test cases.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot:
```java
List<Pair<String, Point>> positions = new ArrayList<>();
positions.add(Pair.of("S1", new Point(100, 100)));
PositionedZoneLayoutFactory factory = new PositionedZoneLayoutFactory(positions);
positions.add(Pair.of("S2", new Point(200, 200))); // would affect factory without copyOf
```
It only avoids the above, but I removed the copy as it was a bit too defensive.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
FredericSabot: Reformulated.

desiredPositions is a user input, so we have no guarantee it does not include overlaps. The end goal is to be able to regenerate diagrams based on saved positions (positions that would be changed in a front-end).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
FredericSabot: Updated the class doc which was wrongly copied from PositionedZoneLayout
`Strictly` denotes that the positions are final.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: Names changed

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: Indeed, I have an infinite loop with
```java
        List<Pair<String, Point>> positions = List.of(
            Pair.of("A", new Point(100, 100)),
            Pair.of("B", new Point(800, 100)),
            Pair.of("C", new Point(500, 100))  // In between A and B with overlaps ==> Overlap resolution will move it elsewhere
        );
```

I fixed this by forcing moves to right/bottom (minimum displacement).

As rectangle positions are fixed one at a time (
```java
for (int i = 1; i < rectangles.size(); i++) {
            boolean moved;
```
), we have the guarantee it will not cause infinite loops.

Typical loops looks like
Rectangle 0: fixed
Rectangle 1: moved right/bottom (position monotically increasing) until does not overlap 0, then fixed
Rectangle 2: moved right/bottom until does not overlap 0 and 1, then fixed
...

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
FredericSabot: Now use
```java
        List<Pair<String, Point>> positions = List.of(
            Pair.of("A", new Point(100, 100)),
            Pair.of("B", new Point(800, 100)),
            Pair.of("C", new Point(500, 100))  // In between A and B with overlaps ==> Overlap resolution will move it elsewhere
        );
```
which was causing an infinite loop before.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
NathanDissoubray: it's fine, since the factory is local

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
NathanDissoubray: By final, do you mean that the object is final, or that the positions do not change ? I don't think `Strictly` is necessary here

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
NathanDissoubray: @FredericSabot Not sure if you did it by accident, but please don't resolve the comments yourself even if the change is simple, that allows me to check

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
NathanDissoubray: if you want to avoid cases where a substation would end very far from its original position, once the move of your rectangle is over, you can check it does not overlap with any rectangle, and if that's the case, move those rectangles in the same direction as the move of the original rectangle (recursively).
That will prevent something like that:
```
0
1 4 (4 overlaps 1, move down)
  2
  3
```
Which gives:
```
0
1
  2
  3
  4 (4 very far from original position)
```

And instead it would be:
```
0
1
  4
  2
  3
```
It's a choice of course, but I think a uniform movement is better for the eye than a substation jumping around

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,151 @@
...
+     * Repeat until no overlaps remain with any earlier substation.
+     */
+    private void resolveOverlaps(List<Rectangle> rectangles, int margin) {
+        for (int i = 1; i < rectangles.size(); i++) {
+            boolean moved;
+            do {
+                moved = false;
+                for (int j = 0; j < i; j++) {
+                    Rectangle fixedRectangle = rectangles.get(j);
+                    Rectangle freeRectangle = rectangles.get(i);
```
NathanDissoubray: how about movingRectangle rather than freeRectangle ?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
FredericSabot: The goal was for this overlap resolution to be as simple as possible (leaving the burden on the user to not input invalid coordinates). On second though, I will check if I can reuse the ForceLayout.

NathanDissoubray: The force layout is not really used for that, I don't think it would work as you expect it to work (since it doesn't take into account shapes / size)
I understand the will to start with something simple, let's keep it like that for now, and we can improve the overlap prevention with another PR.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
NathanDissoubray: Yes I understood that when reading another class, those are questions that other developpers / users could ask. I think it would be good to include the fact that the desired positions can have overlap in the description of the constructor (as well as some javadoc for the other parameters)

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,151 @@
...
+    private final List<Pair<String, Point>> desiredPositions;
+
+    protected PositionedZoneLayout(ZoneGraph graph,
+                                   List<Pair<String, Point>> desiredPositions,
+                                   ZoneLayoutPathFinderFactory pathFinderFactory,
+                                   SubstationLayoutFactory sLayoutFactory,
+                                   VoltageLevelLayoutFactory vLayoutFactory) {
+        super(graph, pathFinderFactory, sLayoutFactory, vLayoutFactory);
+        this.desiredPositions = Objects.requireNonNull(desiredPositions);
+    }
```
NathanDissoubray: Any reason why the constructor of this non-abstract class is protected ?

FredericSabot: Because I am not very fluent in Java and the one of MatrixZoneLayout is. Also, making it private breaks PositionedZoneLayoutFactory

NathanDissoubray: I thought that it could be public actually

FredericSabot: This would allow to use it outside of the factory, I don't know if this is what we want.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
FredericSabot: I think constructor doc is a bit redundant with the (updated) class doc, but I added doc for the desiredPositions parameter (not for the pathFinderFactory, etc. that would be redundant).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: Copy paste as is is not possible as it uses matrix cells instead of positions, but it could indeed be closer to the original code.

I wonder if the diff comes from the additional paths added in empty matrix cells
```java
 private void computeMatrixCellsAvailability(LayoutParameters layoutParameters) {
        // Make empty cells available for snakeline computation
```

I am out next week, but I will investigate this further when I return.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
NathanDissoubray: People might not always need the factory pattern, I don't see a reason to restrain the usage. If people want the factory, they can use the factory. Otherwise, they can use the constructor directly

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
NathanDissoubray: When someone calls the constructor, most tools show the doc of the constructor, not doc of the class, so I don't think that's redundant to explain it there as well.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
```diff
@@ -65,13 +65,14 @@ protected void calculateCoordSubstations(LayoutParameters layoutParameters) {
...
+        int snakeLinePadding = layoutParameters.getZoneLayoutSnakeLinePadding();
         double zoneWidth = getGraph().getSubstations().stream()
             .mapToDouble(sg -> getSubstationOrigin(sg).getX() + sg.getWidth()).max().orElse(0);
         double zoneHeight = getGraph().getSubstations().stream()
             .mapToDouble(sg -> getSubstationOrigin(sg).getY() + sg.getHeight()).max().orElse(0);
         getGraph().setSize(
-            diagramPadding.left() + zoneWidth + diagramPadding.right(),
-            diagramPadding.top() + zoneHeight + diagramPadding.bottom()
+            diagramPadding.left() + zoneWidth + snakeLinePadding + diagramPadding.right(),
+            diagramPadding.top() + zoneHeight + snakeLinePadding + diagramPadding.bottom()
```
NathanDissoubray: I don't think adding `snakeLinePadding` is necessary. The padding of snakelines is meant for spacing the snakelines between themselves.
The padding left / right is meant for the space between the different zones. Increasing spacing between snakelines (which deals with the routing between substations), probably shouldn't move the zones themselves.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -135,27 +136,32 @@ private void computePathFindingGrid(LayoutParameters layoutParameters) {
...
-            int hStep = (int) layoutParameters.getHorizontalSnakeLinePadding();
+            LayoutParameters.Padding vlPadding = layoutParameters.getVoltageLevelPadding();
+            int hStep = Math.max(1, (int) layoutParameters.getHorizontalSnakeLinePadding());
             int zoneW = (int) (width - diagramPadding.left() - diagramPadding.right());
             int startX = (int) diagramPadding.left();
+            int startYTop = ssY - (int) vlPadding.top();
+            int startYBottom = ssY + ssH - (int) vlPadding.bottom();
             for (int x = startX; x < startX + zoneW; x++) {
-                for (int y = ssY - margin; y < ssY; y += Math.max(1, hStep)) {
+                for (int y = startYTop - snakeLinePadding; y < startYTop; y += hStep) {
```
NathanDissoubray, line 150: I'm guessing you substract `snakeLinePadding` here because you added to the `diagramPadding` ? if we remove it from the diagram padding, it should be removed here too.

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
NathanDissoubray: How about `AbstractManuallyPositionedZoneLayout` ?

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
NathanDissoubray: Could you change this to public please ?

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
FredericSabot: I added this since we can have snakelines right of the right most substations. But indeed it can be removed (and the pathFindingGrid clips to zoneWidth anyway).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: Here it is correct: I add a hallway between the top of the substation and the top - snakeLinePadding

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractStrictlyPositionedZoneLayout.java
NathanDissoubray: I think this is a problem overall with the initialisation of the path finding rather than an issue with the positions of the zones. It probably should be solved in a PR on the pathfinding (which I'm working on with #814)

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
NathanDissoubray: But this only works because of `diagramPadding.top() + zoneHeight + snakeLinePadding + diagramPadding.bottom()` right ? If you remove the `+ snakeLinePadding`, then this start could be out of bounds I think.
I also think adding hallways for the pathfinding is out of the scope of this PR

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
```diff
@@ -186,4 +189,26 @@ void testInvalidZoneGraphMatrix() {
...
+        // Run positioned zone layout
+        new PositionedZoneLayoutFactory(positions)
+            .create(g,
+                DijkstraPathFinder::new,
+                new HorizontalSubstationLayoutFactory(),
+                new PositionVoltageLevelLayoutFactory())
+            .run(layoutParameters);
+
+        assertEquals(toString("/TestCase13ZoneGraphPositioned.svg"), toSVG(g, "/TestCase13ZoneGraphPositioned.svg"));
+    }
```
NathanDissoubray, line 257: could you add a test case with more than 1 overlap to resolve, and a test case with no overlap (to make sure there is no overlap prevention happening when not necessary) ?

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: I have made the hallways to start just outside the unavailable defined in
```java
// Mark substation interiors (voltage levels + existing edges) unavailable
...
```

In the html example above, removing the snakeLinePadding from the diagram size only clips the 4th snakeLine

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
FredericSabot: Added same case without the overlap (C at y=600) and full overlap case (everything at (0,0))

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    /** Desired top-left (x, y) per substation, in priority order for overlap resolution. */
+    private final List<Pair<String, Point>> desiredPositions;
+
+    /**
+     * @param graph the zone graph to lay out
+     * @param desiredPositions desired top-left position for each substation, in priority order for overlap resolution;
+     *                         earlier entries win and are never moved
```
NathanDissoubray:
```suggestion
     * @param desiredPositions desired top-left position for each substation. In case of overlap of two substations, 
     *                         the overlap resolution will move the substation with the position of higher index.
```

single-line-diagram/single-line-diagram-core/src/test/resources/TestCase13ZoneGraphPositionedTwoOverlaps.svg
```diff
@@ -0,0 +1,784 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg height="774.0" viewBox="0 0 1430.0 774.0" width="1430.0" xmlns="http://www.w3.org/2000/svg">
+    <style><![CDATA[/* ----------------------------------------------------------------------- */
+/* File: tautologies.css ------------------------------------------------- */
```
NathanDissoubray, line 4: typo in the name of this file

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
```diff
@@ -186,4 +189,70 @@ void testInvalidZoneGraphMatrix() {
...
+
+    @Test
+    void testZoneGraphPositionedNoOverlap() {
+        // build zone graph
+        network = Networks.createNetworkWithManySubstations();
+        List<Pair<String, Point>> positions = List.of(
+            Pair.of("A", new Point(100, 100)),
+            Pair.of("B", new Point(800, 100)),
+            Pair.of("C", new Point(500, 600))
+        );
```
NathanDissoubray, line 201: the resulting SVGs for this test and the oneOverlap test are too similar. Could you put C at something like (100, 600) ?

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/iidm/TestCase13ZoneGraph.java
```diff
@@ -186,4 +189,70 @@ void testInvalidZoneGraphMatrix() {
...
+                DijkstraPathFinder::new,
+                new HorizontalSubstationLayoutFactory(),
+                new PositionVoltageLevelLayoutFactory())
+            .run(layoutParameters);
+
+        assertEquals(toString("/TestCase13ZoneGraphPositionedOneOverlap.svg"), toSVG(g, "/TestCase13ZoneGraphPositionedOneOverlap.svg"));
+    }
+
+    @Test
+    void testZoneGraphPositionedTwoOvelerlaps() {
```
NathanDissoubray:
```suggestion
    void testZoneGraphPositionedTwoOverlaps() {
```

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractManuallyPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+            int startYTop = ssY - (int) vlPadding.top();
+            int startYBottom = ssY + ssH - (int) vlPadding.bottom();
+            for (int x = startX; x < startX + zoneW; x++) {
+                for (int y = startYTop - snakeLinePadding; y < startYTop; y += hStep) {
+                    pathFinderGrid.setAvailability(x, y, true);
+                }
+                for (int y = startYBottom; y <= startYBottom + snakeLinePadding; y += hStep) {
+                    pathFinderGrid.setAvailability(x, y, true);
+                }
+            }
```
NathanDissoubray: could you extract this to a method ? a bit like it was done before in `MatrixZoneLayoutModel`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+            int startXRight = ssX + ssW - (int) vlPadding.right();
+            for (int y = startY; y < startY + zoneH; y++) {
+                for (int x = startXLeft - snakeLinePadding; x < startXLeft; x += vStep) {
+                    pathFinderGrid.setAvailability(x, y, true);
+                }
+                for (int x = startXRight; x <= startXRight + snakeLinePadding; x += vStep) {
+                    pathFinderGrid.setAvailability(x, y, true);
+                }
+            }
+        }
```
NathanDissoubray, line 186: same here for a method

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+                    if (edge instanceof BranchEdge branch) {
+                        Grid.getPointsAlongSnakeline(branch.getSnakeLine())
+                            .forEach(p -> pathFinderGrid.setAvailability(p, false));
+                    }
+                });
+            });
+            sg.getLineEdges().forEach(e ->
+                Grid.getPointsAlongSnakeline(e.getSnakeLine())
+                    .forEach(p -> pathFinderGrid.setAvailability(p, false)));
+        }
```
NathanDissoubray, line 217: and here

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractManuallyPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Zone layout that places substations at desired top-left coordinates (x, y).
+ * <p>
+ * Each substation is assigned a desired top-left position via a {@code List<Pair<String, Point>>}.
+ * Child classes are responsible for running the sub-layouts and making sure the
+ * substations do not overlap (including margin for snakelines).
+ *
+ * @author Frédéric Sabot {@literal <frederic.sabot at haulogy.net>}
+ */
+public abstract class AbstractManuallyPositionedZoneLayout extends AbstractZoneLayout {
```
NathanDissoubray: I know I told you to rename this to `AbstractManuallyPositionedZoneLayout`, but after reading it again, there is actually nothing regarding positioning the zones. I think it should be renamed to something more generic (any suggestion ?), and `PositionedZoneLayout` should be renamed to `ManuallyPositionedZoneLayout`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+ * @author Frédéric Sabot {@literal <frederic.sabot at haulogy.net>}
+ */
+public abstract class AbstractManuallyPositionedZoneLayout extends AbstractZoneLayout {
+
+    /**
+     * Returns the final top-left position for each substation, after running sub-layouts.
+     * Implementations are responsible for calling {@code layoutBySubstation.get(sGraph).run(layoutParameters)}
+     * for each substation before returning.
+     */
+    protected abstract List<Pair<String, Point>> computeSubstationPositions(LayoutParameters layoutParameters);
```
NathanDissoubray, line 39: I'm not a fan of imposing a part of the implementation of this method with just a comment. Either this abstract class should be written in such a way that `layoutBySubstation.get(sGraph).run(layoutParameters)` is always run, and implementing classes do not need to worry about it.
Or this comment should be more generic (talk about what to do, not how to do it). If people need reference, they can always look at classes that implement this abstract class.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,238 @@
...
+        polyline.addAll(pathFinder.findShortestPath(pathFinderGrid, p1, p2));
+        polyline.add(p2);
+        return polyline;
+    }
+
+    // -------------------------------------------------------------------------
+    // Path-finding grid
+    // -------------------------------------------------------------------------
+
+    private void computePathFindingGrid(LayoutParameters layoutParameters) {
```
NathanDissoubray, line 119: this was a public method before in `MatrixZoneLayoutModel`. I'm wondering if it should be protected.
In either case, this is a breaking change, and the description of your pull request will need to be updated to explain migration step (if it's just private, the migration step is just that there is nothing to do).

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/PositionedZoneLayout.java
```diff
@@ -0,0 +1,156 @@
...
+        for (Rectangle r : rectangles) {
+            r.x += snakeLinePadding - minX;
+            r.y += snakeLinePadding - minY;
+        }
+
+        List<Pair<String, Point>> positions = new ArrayList<>();
+        for (int i = 0; i < substationGraphs.size(); i++) {
+            Rectangle r = rectangles.get(i);
+            positions.add(Pair.of(substationGraphs.get(i).getId(), new Point(r.x, r.y)));
+        }
```
NathanDissoubray: these two for-loops could be merged into a single one

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
NathanDissoubray: After checking your PR description, I see you removed a part of the template regarding breaking changes. Could you add it back please ? You can see the template in other PR such as https://github.com/powsybl/powsybl-diagram/pull/820 or here https://github.com/powsybl/.github/blob/main/PULL_REQUEST_TEMPLATE.md

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractManuallyPositionedZoneLayout.java
FredericSabot: computeSubstationPositions is about positions.
Changed
AbstractManuallyPositionedZoneLayout -> AbstractPositionedZoneLayout
PositionedZoneLayout -> ManuallyPositionedZoneLayout

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: The contract is that implementations return zone positions and this compute the pathfinding grid around, so I don't see a need for protected. Especially, given the we want to delete this function anyway once we have pathfinding.

FredericSabot: Added migration steps in PR.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: Moved to call to the base class and updated the doc accordingly.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
NathanDissoubray: I think in the case where a method is removed and there is no replacement, it is good to specify in the migration steps why this was done, and what is the impact. For example, here there should be no impact, since the code is still called privately by another method.

NathanDissoubray (dismissed):
I'm still not too enthusiastic about this PR changing the pathfinding slightly, by changing the padding of some areas.
However, it need works anyway, and it'll be changed with #814 anyway. I see why those changes were made when moving code, and I feel like it's slightly better that way.

I'll approve, but I cannot merge this as I'm not a maintainer, you will have to wait for another review @FredericSabot

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
FredericSabot: Updated

NathanDissoubray:
@FredericSabot Could you fix the issue reported by Sonar please ?

NathanDissoubray (approved):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/AbstractPositionedZoneLayout.java
```diff
@@ -0,0 +1,259 @@
...
+import com.powsybl.sld.model.nodes.BranchEdge;
+import com.powsybl.sld.model.nodes.Node;
+import org.jgrapht.alg.util.Pair;
+
+import java.util.*;
+
+/**
+ * Zone layout that places substations at desired top-left coordinates (x, y).
+ * <p>
+ * Each substation is assigned a desired top-left position via a {@code List<Pair<String, Point>>}.
```
NathanDissoubray: I missed something in my previous review

I don't think "desired" is the correct wording here. We only "desire" a position in `ManuallyPositionnedZoneLayout` (whereas in `MatrixZoneLayout`, those positions are calculated based on a matrix provided by the user).
I think it could be changed to something more like "Position substations given input from the user".
Yes the positions are given by `protected abstract List<Pair<String, Point>> computeSubstationPositions(LayoutParameters layoutParameters)`, but that's an implementation detail, and I don't think this should be in the description of the class.

NathanDissoubray (changes requested):
I know I already approved, but I missed some stuff, sorry about that

So-Fras (changes requested):
Thank you for your work @FredericSabot! Could you please add some documentation on your work by Friday morning? 
See: https://powsybl.readthedocs.io/projects/powsybl-diagram/latest/single_line_diagrams/layout/zoneMatrixLayout.html

docs/single_line_diagrams/layout/zoneLayouts.md
```diff
@@ -0,0 +1,122 @@
...
+## Input parameters
+
+- `VoltageLevelLayoutFactory`: builder of the layout used by voltage levels
+- `SubstationLayoutFactory`: builder of the layout used by substations
+- `ZoneLayoutPathFinderFactory`: builder of the pathfinders used to draw lines between substations
+- `desiredPositions`: list of the substations and their desired positions (ex: `List.of(Pair.of("A", new Point(100, 100)), Pair.of("B", new Point(800, 100)), Pair.of("C", new Point(500, 100)))`)
+
+**Usage example:**
+The following example displays three substations distributed on two columns and two lines,
+with an empty area at the middle of the second line.
```
NathanDissoubray: Update this for the current use case. Can you add an image that corresponds to this test in the documation ? Maybe by higlighting the A, B and C substations (a simple rectangle of different color for A, B and C would workd I think)

docs/single_line_diagrams/layout/zoneMatrixLayout.md
```diff
@@ -1,123 +0,0 @@
...
-|                 |                     |    __ZonePadding__    |                      |                 |                     |    __ZonePadding__    |                      |                 |
-
-
-The class `MatrixZoneLayout` represents the matrix layout.<BR>
-The class `MatrixZoneLayoutModel` represents the matrix and the path finder information.
-The class `Matrix` contains an array of `MatrixCell`.
-
-The class `MatrixCell` stores information related to the matrix cell:
-- Position (indices) in the matrix: row, column
-- The id of the substation graph contained by the cell
```
NathanDissoubray, line 63: This was removed, but this is still relevant for the matrix layout. Could you add it back please ?

docs/single_line_diagrams/layout/zoneLayouts.md
```diff
@@ -0,0 +1,122 @@
...
+getGraph().getSubstations().forEach(sg -> layoutBySubstation.get(sg).run(layoutParameters));
+```
+
+This allows to get the size of each substation diagram. Then, it calls the abstract method `computeSubstationPositions`. Implementations of this method are responsible to compute substation positions that do not cause overlaps between substations, including a margin for `layoutParameters.getZoneLayoutSnakeLinePadding()`.
+
+For the `MatrixZoneLayout`, this is done by placing substations in a matrix, where the height of each row is taken as the height of its largest element plus the snakeline margin (and the same for the width of each column).
+
+For the `ManuallyPositionedZoneLayout`, this is done by placing substations at user-defined positions, then automatically resolving overlaps.
+
+Once those substations positions are computed, the `AbstractPositionedZoneLayout` move the substations at those positions and then size the diagram accordingly
```
NathanDissoubray:
```suggestion
Once those substations positions are computed, the `AbstractPositionedZoneLayout` moves the substations at those positions and then scales the diagram accordingly
```

NathanDissoubray (changes requested):

docs/single_line_diagrams/layout/zoneMatrixLayout.md
FredericSabot: The MatrixZoneLayoutModel and associated Matrix, MatrixCell have been removed.

NathanDissoubray:
@FredericSabot you updated the doc for Matrix Zone Layout (which is good, and it fixes a mistake in there, so thank you for that). However, the doc for the manually positionned zone layout remains the same (the description for the usage example is incorrect, and there is no image). The example for the padding is still missing from the manually positionned zone layout.

I can update the doc if you want, or you can do it, whichever you prefer.

docs/single_line_diagrams/layout/zoneMatrixLayout.md
NathanDissoubray: no, they are still used by the MatrixZoneLayout and are still present

NathanDissoubray:
@FredericSabot I made a PR on your branch (on your fork), so if you are fine with the changes I suggested, you can merge it there and it should appear as a commit in this pull request (hopefully)

So-Fras (approved):

NathanDissoubray (approved):
@FredericSabot thank you for the contribution !

PR #820, opened 2026-06-11
Multiple selected limits + prevent throw on low limit

NathanDissoubray:

Take into account when multiple limits are selected.
Avoid throwing error when permanent limit is queried from a current limit that is a low limit.

Only the last selected limit is taken into account.
If the low limit is the last selected limit, getting feeder info will throw an error, because there is no permanent limit for a low limit.

All selected limits are used.
Check that there is a permanent limit before getting it. If we have a low limit, return the default value instead.

- [ ] Yes
- [x] No

**Other information**:

Needs core 7.3.0 to work for the low limits


NathanDissoubray:
I don't think I can increase coverage anymore than this. It would require to have a null `CurrentLimits`, and I don't think that's possible since:
- the methods that gives `CurrentLimits` to `getPermanentLimitPercentageMax` do so by calling `getAllSelectedCurrentLimits().stream()`
- to get something null in the stream, it would require a null limit in a selected group
- limits are created by an adder and cannot be null.
- if no current limit exists, `getCurrentLimits` returns an empty optional, `getAllSelectedCurrentLimits` would then return an empty collection.

So with this, I don't think it's possible to have a null currentLimits given to `getPermanentLimitPercentageMax`

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultLabelProvider.java
```diff
@@ -200,18 +200,22 @@ private List<FeederInfo> get3WTFeederInfos(ThreeWindingsTransformer transformer,
 
     private double getPermanentLimitPercentageMax(Branch<?> branch) {
         return Stream.of(TwoSides.ONE, TwoSides.TWO)
-            .map(side -> getPermanentLimitPercentageMax(branch.getTerminal(side), branch.getCurrentLimits(side).orElse(null)))
-            .mapToDouble(Double::doubleValue).max().getAsDouble();
+            .flatMap(side -> branch.getAllSelectedCurrentLimits(side).stream()
```
So-Fras: The 3 functions are present in the PermanentLimitPercentageMax class. We should re-use them!

Beside, returning 0 when there is no limit is weird. We should do what is done in NAD and return NaN.

So-Fras (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/DefaultLabelProvider.java
NathanDissoubray: Agreed, done in 4ab045200549

So-Fras (approved):

PR #829, opened 2026-08-04
fix Null network in diagram style provider

### LINKED ISSUE #181: Null network in diagram style provider
Bug

Network can be null in `AbstractBaseVoltageDiagramStyleProvider`, leading to NullPointerExceptions.

    - Provide a null network in `TopologicalStyleProvider` or `NominalVoltageDiagramStyleProvider`
    - Do not disable highlightLineState in `LayoutParameters`  
    - Enable addNodesInfos: `LayoutParameters.setAddNodesInfos(true)`

No NPE. Or exception when creating DiagramStyleProvider, but I think some users which don't handle a network still need a NominalVoltageDiagramStyleProvider (see this spectrum [question](https://spectrum.chat/powsybl/issues/release-sld-1-6-0-missing-constructor-nominalvoltagediagramstyleprovider~c69b1c2f-6156-430f-bb76-239f568b1549)).

Avoid NPE.

samirromdhani:

Fixes #181

- [x] Yes
- [ ] No

`HighlightLineStateStyleProvider(Network)` now rejects a `null` network at construction time.
Pass a non-null `Network` to `HighlightLineStateStyleProvider`.

**Other information**:

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
@@ -31,11 +31,7 @@ import com.powsybl.sld.model.nodes.feeders.FeederWithSides;
 import com.powsybl.sld.svg.styles.EmptyStyleProvider;
 import com.powsybl.sld.svg.styles.StyleClassConstants;
 
-import java.util.Collections;
-import java.util.EnumMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 
 /**
  * @author Franck Lecuyer {@literal <franck.lecuyer at rte-france.com>}
@@ -46,6 +42,7 @@ public class HighlightLineStateStyleProvider extends EmptyStyleProvider {
     private final Network network;
 
     public HighlightLineStateStyleProvider(Network network) {
+        Objects.requireNonNull(network);
         this.network = network;
     }
 
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
@@ -41,6 +41,10 @@ public class HighlightLineStateStyleProvider extends EmptyStyleProvider {
 
     private final Network network;
 
+    /**
+     *
+     * @param network the IIDM network (must not be {@code null})
+     */
     public HighlightLineStateStyleProvider(Network network) {
         Objects.requireNonNull(network);
         this.network = network;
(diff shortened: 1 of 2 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
```diff
@@ -46,6 +42,7 @@ public class HighlightLineStateStyleProvider extends EmptyStyleProvider {
     private final Network network;
 
     public HighlightLineStateStyleProvider(Network network) {
+        Objects.requireNonNull(network);
```
NathanDissoubray, line 49: I think this should be marked as a breaking change, and the function javadoc description should include this requirement

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
samirromdhani: Added javadoc and marked the PR as a breaking change

NathanDissoubray (approved):
When reading the original issue, it states that the goal it to avoid an NPE. `Objects.requireNonNull` will still launch an NPE, but I think this is fine, since I think what @flo-dup meant was to avoid an NPE when launching another function :  the source of the error should originate from the function where the error can be fixed, which is fixed by this PR.
So it will still launch an NPE, but in a place where the error can be corrected.

samirromdhani:
> When reading the original issue, it states that the goal it to avoid an NPE. `Objects.requireNonNull` will still launch an NPE, but I think this is fine, since I think what @flo-dup meant was to avoid an NPE when launching another function : the source of the error should originate from the function where the error can be fixed, which is fixed by this PR. So it will still launch an NPE, but in a place where the error can be corrected.

Yes, `Objects.requireNonNull `still throws an NPE, but the point is to fail at construction time rather than deep inside another method.  The NPE can be from any method in the `HighlightLineStateStyleProvider` object, same conclusion documented in the original issue https://github.com/powsybl/powsybl-diagram/issues/181#issuecomment-5176988213 before opening this PR.

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/util/TopologicalStyleTest.java
```diff
@@ -191,4 +192,15 @@ void testBusHighlightSubstationUnifiedVoltageLevelColors() {
                 new DefaultSVGLegendWriter(network, svgParameters)));
     }
 
+    @Test
+    void defaultStyleProviderThrowsWhenNetworkIsNull() {
+        network = null;
+        assertThatCode(this::getDefaultDiagramStyleProvider).isInstanceOf(NullPointerException.class);
```
So-Fras, line 198: Why using AssertThatCode instead of AssertThrows (and AssertDoesNotThrow)?
If there is no difference, then I suggest to keep AssertThrows / AssertDoesNotThrow for consistency.

So-Fras (changes requested):

PR #830, opened 2026-08-04
Fix non-idempotent voltage level layout run

### LINKED ISSUE #289: Running the layout is not idempotent
Bug

Running the layout is not idempotent

Running twice the layout on the same Diagram object leads to ugly glitches (wider cells, floating internal nodes, margins between voltage levels wider, ....)

Running twice the layout on the same Diagram object leads to the same SVG file

Fixing a bug

 PowSyBl Version: core 4.4.0 and sld 2.4.0

samirromdhani:

Fixes #289 

- [ ] Yes
- [ ] No

**Other information**:

##### Root cause

- `cellDetector.detectCells(getGraph()); `always creates new cells via graph.addCell()
- `addPaddingToCoord() `add vl padding to graph.coord on every run (coord is not reset)

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
@@ -592,14 +592,14 @@ public class VoltageLevelGraph extends AbstractBaseGraph {
     }
 
     public Point getCoord() {
-        return coord;
+        return coord;// new Point(0, 0);
     }
 
     public void addPaddingToCoord(LayoutParameters layoutParam) {
         LayoutParameters.Padding vlPadding = layoutParam.getVoltageLevelPadding();
         LayoutParameters.Padding dPadding = layoutParam.getDiagramPadding();
-        setCoord(coord.getX() + dPadding.left() + vlPadding.left(),
-                coord.getY() + dPadding.top() + vlPadding.top());
+        setCoord(getCoord().getX() + dPadding.left() + vlPadding.left(),
+                getCoord().getY() + dPadding.top() + vlPadding.top());
     }
 
     public void setCoord(double x, double y) {
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
@@ -592,7 +592,7 @@ public class VoltageLevelGraph extends AbstractBaseGraph {
     }
 
     public Point getCoord() {
-        return coord;// new Point(0, 0);
+        return coord;
     }
 
     public void addPaddingToCoord(LayoutParameters layoutParam) {
@@ -781,5 +781,11 @@ public class VoltageLevelGraph extends AbstractBaseGraph {
     }
 
     private record GroundDisconnection(List<Node> nodes, FeederNode ground, SwitchNode disconnector, Node forkNode) {
+    }
+
+    public void clearCells() {
+        cells.clear();
+        cellCounter = 0;
+        setCoord(0, 0);
     }
 }
--- a/single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/LayoutIdempotencyTest.java
+++ b/single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/LayoutIdempotencyTest.java
@@ -42,16 +42,24 @@ class LayoutIdempotencyTest extends AbstractTestCaseIidm {
     }
 
     @Test
-    void layoutRunTwiceShouldProduceSameGraph() {
-        // Given
+    void layoutRunTwiceShouldProduceSameGraphPositionFactory() {
+        assertLayoutIdempotent(new PositionVoltageLevelLayoutFactory());
+    }
+
+    @Test
+    void layoutRunTwiceShouldProduceSameGraphSmartFactory() {
+        assertLayoutIdempotent(new SmartVoltageLevelLayoutFactory(network));
+    }
+
+    void assertLayoutIdempotent(VoltageLevelLayoutFactory layoutFactory) {
+        // Given (layout)
         VoltageLevelGraph graph = graphBuilder.buildVoltageLevelGraph(vl.getId());
-        Layout layout = new PositionVoltageLevelLayoutFactory().create(graph);
-//        var layout = new SmartVoltageLevelLayoutFactory(network).create(graph);
-        // When run (first call)
+        Layout layout = layoutFactory.create(graph);
+        // When run Layout (first call)
         layout.run(layoutParameter);
         String afterFirstRun = toJson(graph, "/afterFirstRun.json");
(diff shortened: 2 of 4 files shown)


NathanDissoubray (commented):
Have you checked that there isn't the same kind of bug for `ZoneGraph` and for `SubstationGraph` ?

samirromdhani:
> Have you checked that there isn't the same kind of bug for `ZoneGraph` and for `SubstationGraph` ?

Yes, both use `VoltageLevelLayoutFactory`, so the fix applies there too.
Added equivalent tests for substation and zone graph.

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/raw/TestCase7CellDetectionIssue.java
```diff
@@ -42,4 +42,26 @@ void test() {
...
+    void detectCellsTwiceWithClearIsIdempotent() {
+        VoltageLevelGraph g = rawGraphBuilder.buildVoltageLevelGraph("vl");
+        ImplicitCellDetector cellDetector = new ImplicitCellDetector();
+        cellDetector.detectCells(g);
+        assertEquals(1, g.getCellStream().count());
+
+        g.clearCells();
+        cellDetector.detectCells(g);
+        assertEquals(1, g.getCellStream().count()); // fix bug 289
+    }
```
NathanDissoubray: I'm not sure if those tests are necessary, will people use the raw graph instead of the layouts ?
and even if that's the case, I don't think the first test that checks the bug is necessary

samirromdhani: Following the test class: `CellDetectionIssue`, I considered it the right place to unit test the cell detection behavior, regardless of cell detection usage.

NathanDissoubray: What I mean is that writing `detectCellsTwiceReturnDuplicatesCells` means we expect that running the detection twice will double the number of cells. In my opinion, this is actually an undefined behaviour, do we want to say through this test that this is what we expect ?

The second test can stay as it checks that clearing makes everything work correctly

samirromdhani: Ok. The second test is enough.

NathanDissoubray (dismissed):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
```diff
@@ -598,8 +598,8 @@ public Point getCoord() {
     public void addPaddingToCoord(LayoutParameters layoutParam) {
         LayoutParameters.Padding vlPadding = layoutParam.getVoltageLevelPadding();
         LayoutParameters.Padding dPadding = layoutParam.getDiagramPadding();
-        setCoord(coord.getX() + dPadding.left() + vlPadding.left(),
-                coord.getY() + dPadding.top() + vlPadding.top());
+        setCoord(getCoord().getX() + dPadding.left() + vlPadding.left(),
```
So-Fras, line 601: I don't see the link with the 2nd root cause mentioned in the PR description.

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/LayoutIdempotencyTest.java
```diff
@@ -0,0 +1,90 @@
...
+    LayoutParameters layoutParameter = new LayoutParameters();
+
+    @BeforeEach
+    public void setUp() {
+        network = Network.create("testCase1", "test");
+        graphBuilder = new NetworkGraphBuilder(network);
+        substation = Networks.createSubstation(network, "s", "s", Country.FR);
+        vl = Networks.createVoltageLevel(substation, "vl", "vl", TopologyKind.NODE_BREAKER, 380);
+        Networks.createBusBarSection(vl, "bbs", "bbs", 0, 1, 1);
+        Networks.createLoad(vl, "l", "l", "l", 0, ConnectablePosition.Direction.TOP, 2, 10, 10);
```
So-Fras:
```suggestion
        Networks.createLoad(vl, "l", 0, ConnectablePosition.Direction.TOP, 2, 10, 10);
```

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/LayoutIdempotencyTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    LayoutParameters layoutParameter = new LayoutParameters();
+
+    @BeforeEach
+    public void setUp() {
+        network = Network.create("testCase1", "test");
+        graphBuilder = new NetworkGraphBuilder(network);
+        substation = Networks.createSubstation(network, "s", "s", Country.FR);
+        vl = Networks.createVoltageLevel(substation, "vl", "vl", TopologyKind.NODE_BREAKER, 380);
+        Networks.createBusBarSection(vl, "bbs", "bbs", 0, 1, 1);
```
So-Fras:
```suggestion
        Networks.createBusBarSection(vl, "bbs", 0, 1, 1);
```

So-Fras (changes requested):
Thanks for the work! I left first comments to discuss.

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java
samirromdhani: Here, I just preferred using the getter, the two root causes are added in the same method `clearCells`

```java
    public void clearCells() {
        cells.clear();        // first root cause
        cellCounter = 0;      // first root cause
        setCoord(0, 0);       // second root cause
    }
```

samirromdhani: The root causes:

https://github.com/powsybl/powsybl-diagram/blob/952186b5b34d1e4e472a04fb663b3654b54e3022/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/layout/ImplicitCellDetector.java#L67-L68
https://github.com/powsybl/powsybl-diagram/blob/952186b5b34d1e4e472a04fb663b3654b54e3022/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java#L496-L497

`cells` was not initialized when the layout was run twice =>  fixed by `cells.clear()`
`graph.getNextCellNumber()` was not initialized when the layout was run twice  =>  fixed by `cellCounter = 0`

https://github.com/powsybl/powsybl-diagram/blob/952186b5b34d1e4e472a04fb663b3654b54e3022/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/model/graphs/VoltageLevelGraph.java#L601

`Coord` was not initialized when the layout was run twice =>  fixed by `setCoord(0, 0)`

NathanDissoubray (approved):

So-Fras (approved):

PR #831, opened 2026-08-11
fix HighlightLineState feeder bug

### LINKED ISSUE #251: Fix HighlightLineState feeder bug
Bug

HighlightLineState only applies to the FeederPrimaryBlock, it stops at its fictitious node
highlightLineStateFeeder

Use NominalVoltageDiagramStyleProvider or TopologicalStyleProvider with LayoutParameters::highlightLineState set to true (default value).

Style applied from feeder to the first component

samirromdhani:

Fixes #251

Bug fix

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
+++ b/single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
@@ -35,6 +35,7 @@ import java.util.Collections;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 /**
@@ -93,8 +94,34 @@ public class HighlightLineStateStyleProvider extends EmptyStyleProvider {
         } else if (n1 instanceof Internal2WTNode || n2 instanceof Internal2WTNode) {
             return getHighlightFeederStateStyleForInternal2WT(n1, n2);
         } else {
+            return getFeederBlockEdgeStyle(graph, edge);
+        }
+    }
+
+    private Optional<String> getFeederBlockEdgeStyle(Graph graph, Edge edge) {
+        Node n1 = edge.getNode1();
+        Node n2 = edge.getNode2();
+        Node fictitious;
+        if (n1.isFictitious()) {
+            fictitious = n1;
+        } else if (n2.isFictitious()) {
+            fictitious = n2;
+        } else {
+            fictitious = null;
+        }
+        if (fictitious == null) {
             return Optional.empty();
         }
+        Node adjacentNode = fictitious == n1 ? n2 : n1;
+        if (adjacentNode instanceof FeederNode || adjacentNode.isFictitious()) {
+            return Optional.empty();
+        }
+        return fictitious.getAdjacentNodes().stream()
+                .filter(FeederNode.class::isInstance)
+                .map(FeederNode.class::cast)
+                .filter(feeder -> !Objects.equals(adjacentNode.getId(), feeder.getId()))
+                .findFirst()
+                .flatMap(feeder -> getHighlightFeederStateStyle(graph, feeder));
     }
 
     protected Optional<String> getHighlightFeederStateStyle(Graph graph, FeederNode n) {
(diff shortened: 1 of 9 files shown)


single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
```diff
@@ -93,8 +93,33 @@ private Optional<String> getHighlightLineStateStyle(Graph graph, Edge edge) {
...
+        if (n1.isFictitious()) {
+            fictitious = n1;
+        } else if (n2.isFictitious()) {
+            fictitious = n2;
+        } else {
+            fictitious = null;
+        }
+        if (fictitious == null) {
+            return Optional.empty();
+        }
```
NathanDissoubray, line 110:
```suggestion
        Node n1 = edge.getNode1();
        Node n2 = edge.getNode2();
        Node fictitious;
        if (n1.isFictitious()) {
            fictitious = n1;
        } else if (n2.isFictitious()) {
            fictitious = n2;
        } else {
            return Optional.empty()
        }
```

NathanDissoubray: Since you are using `node.isFictitious()`, I assume that means that node1 or 2 cannot be null.
The check of `(fictitious == null)` after could be make sense if there was more than one assignement that could make `fictitious` null, but since `n1` and `n2` can't be null (given the call to `isFictitious`), there is only the "else" where it is possible (so you can directly return).

If n1 or n2 can be null, then there needs to be a check for that before calling `.isFictitious`

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/util/HighlightLineStateStyleTest.java
```diff
@@ -0,0 +1,64 @@
...
+        voltageLevelGraphLayout(graph);
+        StyleProviderFactory styleFactory = new DefaultStyleProviderFactory(); // HighlightLineStateStyleProvider included by default
+        StyleProvider styleProvider = styleFactory.create(network, svgParameters);
+        // Then
+        assertThat(graph.getEdges())
+                .filteredOn(edge -> styleProvider.getEdgeStyles(graph, edge)
+                        .contains(StyleClassConstants.FEEDER_DISCONNECTED_CONNECTED))
+                .hasSize(2)
+                .extracting(edge -> edge.getNodes().stream().map(Node::getId).toList())
+                .containsExactlyInAnyOrder(List.of("INTERNAL_vl1_load", "load"), List.of("breaker", "INTERNAL_vl1_load"));
```
NathanDissoubray, line 62: so if I understand correctly, you are checking that both the edge from the internal node to the load, and the edge from the breaker to the internal node have the disconnected style ?

NathanDissoubray (changes requested):

single-line-diagram/single-line-diagram-core/src/test/java/com/powsybl/sld/util/HighlightLineStateStyleTest.java
samirromdhani: Yes, correct

NathanDissoubray (approved):

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
```diff
@@ -93,8 +93,30 @@ private Optional<String> getHighlightLineStateStyle(Graph graph, Edge edge) {
...
+        Node n2 = edge.getNode2();
+        Node fictitious;
+        if (n1.isFictitious()) {
+            fictitious = n1;
+        } else if (n2.isFictitious()) {
+            fictitious = n2;
+        } else {
+            return Optional.empty();
+        }
+        Node adjacentNode = fictitious == n1 ? n2 : n1;
```
Hyralc, line 111: This code actually already exists somewhere, maybe we could reuse it:

```suggestion
        Node adjacentNode = edge.getOppositeNode(fictitious);
```

single-line-diagram/single-line-diagram-core/src/main/java/com/powsybl/sld/svg/styles/iidm/HighlightLineStateStyleProvider.java
```diff
@@ -93,8 +93,30 @@ private Optional<String> getHighlightLineStateStyle(Graph graph, Edge edge) {
...
+        Node n1 = edge.getNode1();
+        Node n2 = edge.getNode2();
+        Node fictitious;
+        if (n1.isFictitious()) {
+            fictitious = n1;
+        } else if (n2.isFictitious()) {
+            fictitious = n2;
+        } else {
+            return Optional.empty();
+        }
```
Hyralc, line 110: In order to match with the end of the method, we could do this:

```suggestion
    Node fictitious = edge.getNodes().stream()
            .filter(Node::isFictitious)
            .findFirst()
            .orElse(null);

    if (fictitiousNode == null) {
        return Optional.empty();
    }
```

PR #833, opened 2026-08-18
[NAD] Customize injection style

### LINKED ISSUE #717: Add injections to CustomStyleProvider

It is not possible to customize injection style through the CustomStyleProvider

Users should be able to customize injection style through the CustomStyleProvider

### Describe the motivation

Better customization

### Extra Information

_No response_

samirromdhani:

Fixes #717 

- [ ] Yes
- [x] No

**Other information**:

<p align="center"> before - after</p>
<p align="center">

svg-before

svg-after

svg-after

</p>


network-area-diagram/src/test/java/com/powsybl/nad/svg/CustomStyleProviderTest.java
```diff
@@ -74,16 +94,11 @@ void testCustomStyleProvider() {
...
-
-    @Test
-    void testCustomStyleProviderEmpty() {
-        Network network = Networks.createNodeBreakerNetworkWithBranchStatus("TestNodeDecorators", "test");
-        styleProvider = new CustomStyleProvider(new HashMap<>(), new HashMap<>(), new HashMap<>());
-        assertSvgEquals("/custom_style_provider_empty.svg", network);
+        Map<String, InjectionStyles> injectionStyles = new HashMap<>();
+        injectionStyles.put("G", new InjectionStyles("SteelBlue", "4px", null));
+        injectionStyles.put("L1", new InjectionStyles("SteelBlue", "4px", null));
+        injectionStyles.put("L2", new InjectionStyles("SteelBlue", "4px", "1"));
```
NathanDissoubray: could you use different values for the color and the stroke width ? and maybe test one more dash value, like 10 or 20

network-area-diagram/src/main/java/com/powsybl/nad/svg/CustomStyleProvider.java
```diff
@@ -60,11 +66,15 @@ public record ThreeWtStyles(String edge1, String width1, String dash1, String ed
     private record EdgeStyle(String stroke, String strokeWidth, String dash) {
     }
 
+    public record InjectionStyles(String stroke, String strokeWidth, String dash) {
```
NathanDissoubray: could add some javadoc to give an example of the values ? you could just copy one of the test `new InjectionStyles("SteelBlue", "4px", "1")`

NathanDissoubray (changes requested):

network-area-diagram/src/test/java/com/powsybl/nad/svg/CustomStyleProviderTest.java
samirromdhani: done

network-area-diagram/src/main/java/com/powsybl/nad/svg/CustomStyleProvider.java
samirromdhani: done

NathanDissoubray (commented):
Image

Image

I'm wondering if the dash should also apply inside the symbol of the injection, and not just on the line 🤔

samirromdhani:
> I'm wondering if the dash should also apply inside the symbol of the injection, and not just on the line 🤔

From what I can see this is by design, the style is applied before constructing the injection's elements, 

https://github.com/powsybl/powsybl-diagram/blob/ab67c4bf78f54824b32a074f82257d6c7121d7c7/network-area-diagram/src/main/java/com/powsybl/nad/svg/SvgWriter.java#L190-L193

In SvgWriter, `writeStyleAttribute` applies the style first, then

`drawInjectionEdge`   -> draw the line
`drawInjectionIcon`    -> draw the circle including the symbol of the injection within the `insertSvgComponent` method.
So the style is applied to all the elements composing the injection.

NathanDissoubray:
Since drawing the edge and the icon are separate, doesn't that imply that we could ignore the style of the edge for the icon of the injection ?

samirromdhani:
> Since drawing the edge and the icon are separate, doesn't that imply that we could ignore the style of the edge for the icon of the injection ?

I don't see that as an implication.
I did a quick test by applying the style only to the edge
- remove `writeStyleAttribute` from `drawInjections` and add it to `drawInjectionEdge`

At first impression, the injection style looks incomplete.
image

NathanDissoubray:
I would keep the color of the edge for the icon of the injection, but I don't know for the dash, it looks strange, especially on the blue one for example. It's unclear that it is a generator, so in the case of the blue generator, does that mean I need to choose the style of the edge depending on the symbol of the injection itself ?

samirromdhani:
The blue one is a load, not a generator.
The icons are also SVG, injected through CSS classes. You can check `components.json`,

